PR #17, opened 2017-05-14
add contributing guide

retokromer:


ablwr:
Shouldn't this direct people to the IETF mailing list, and maybe also the FLAC mailing list?

retokromer:
Good point!

- [ ] [EBML](https://github.com/Matroska-Org/ebml-specification)
- [x] [Matroska]()

privatezero:
I also see that it is now very easy to add a CODE_OF_CONDUCT file, would other's be in favor of adding one?

dericed:
The IETF Code of Conduct would be appropriate. For other code of conducts you should probably consult to working group chairs or gather consensus.

retokromer:
I personally have the Code of Conduct blues. The oldest I know has been released by… the mafia in Sicily! Yet I second the proposition of the IETF’s one.

retokromer:
@dericed Do you mean [this one](https://www.internetsociety.org/get-involved/join-community/individuals/code-conduct)?

privatezero:
Possibly [this one?](https://tools.ietf.org/html/rfc7154) That seems to be the most current one they have published.

When it comes to articulated codes, I suppose my feeling is, best to have one and not need it than vice versa!

retokromer:
Yes, I guess this is what Dave was referring to!

I agree with you in theory, but the problem I have often noticed, in practice, is the difficulty to act when one feels the CoC has been broken.

PR #19, opened 2017-05-15
Bullet lists

privatezero:
Fixes rendering of lists in derivatives


retokromer:
LGTM

flac.md
```diff
@@ -237,7 +237,7 @@ NOTES
   - 14 - During recording
   - 15 - During performance
   - 16 - Movie/video screen capture
-  - 17 - A bright coloured fish
+  - 17 - A bright colored fish
```
privatezero: Sorry to be extremely pedantic, but the British spelling 'coloured' seems to be the spelling used not just here, but also in ID3 and most likely is not a true typo. Are we standardizing to American English  spellings?

retokromer: It’s fine to be pedantic! All the other classics I noticed (e.g. -ise vs. -ize) are American English here, therefore I think it would be reasonable to unify to it.

(Charles Poynton always uses colour, coloured ;-)

privatezero: Sounds good!

Issue #20, opened 2017-05-15
list hierarchy mismatch


dericed:
The line `The Rice partition order in a Rice-coded residual section must be less than or equal to 8.` is at a different level than it exists in https://xiph.org/flac/format.html.

retokromer:
Thank you!

PR #21, opened 2017-05-15
fix formatting

retokromer:
should resolve https://github.com/privatezero/flac_markdown/issues/20


retokromer:
should resolve https://github.com/privatezero/flac_markdown/issues/22

privatezero:
Looks good!

PR #24, opened 2017-05-15
Format cross references

privatezero:
Reworded so that anchors make sense in both the md version and the post makefile version.

During this process, noticed that there are some issues with the internal markdown links behaving properly - will open an issue!


privatezero:
oops - looks like i messed up clearing the conflicts! will do new pull.

PR #35, opened 2017-05-23
Improve links

privatezero:
* Changes links to render better in rfc
* Fixes link to registration page


flac.md
```diff
@@ -113,11 +113,11 @@ Before the formal description of the stream, an overview might be helpful.
...
-  - The blocksize bits in the [frame header](#frameheader) must be 0001-1110. The blocksize must be <= 16384; if the sample rate is <= 48000 Hz, the blocksize must be <= 4608.
-  - The sample rate bits in the [frame header](#frameheader) must be 0001-1110.
-  - The bits-per-sample bits in the [frame header](#frameheader) must be 001-111.
-  - If the sample rate is <= 48000 Hz, the filter order in [LPC subframes](#subframelpc) must be less than or equal to 12, i.e. the subframe type bits in the [subframe header](#subframeheader) may not be 101100-111111.
-   - The Rice partition order in a [Rice-coded residual section](#residualcodingmethodpartitionedrice) must be less than or equal to 8.
+  - The blocksize bits in the `frame header` (see [section on `frame header`](#frameheader)) must be 0001-1110. The blocksize must be <= 16384; if the sample rate is <= 48000 Hz, the blocksize must be <= 4608.
```
dericed: Shouldn't `frame header` be in all caps?

privatezero: Would that apply to capitalization for all named anchors?  I left them as is in original doc, but debated capitalizing them all.

privatezero:
Any objections merging this in current form?

retokromer:
I personally have the impression that @dericed means `FRAME_HEADER` and not of `Frame Header`.

privatezero:
Ah, that makes sense! I'll make changes tomorrow to reflect both natural language and the form of the anchor in MD for all links.

dericed:
@retokromer, though it's a bigger issue that the concept of `frame header` and `FRAME_HEADER` appear identical but use differing case. I'll file an issue.

privatezero:
made some changes to reflect https://github.com/privatezero/flac_markdown/issues/43
I'll look at conflicts once other larger pulls are merged!

flac.md
```diff
@@ -113,11 +113,11 @@ Before the formal description of the stream, an overview might be helpful.
...
-  - The blocksize bits in the `frame header` (see [section on `frame header`](#frameheader)) must be 0001-1110. The blocksize must be <= 16384; if the sample rate is <= 48000 Hz, the blocksize must be <= 4608.
-  - The sample rate bits in the `frame header` (see [section on `frame header`](#frameheader)) must be 0001-1110.
-  - The bits-per-sample bits in the `frame header` (see [section on `frame header`](#frameheader))  must be 001-111.
-  - If the sample rate is <= 48000 Hz, the filter order in `LPC subframes` (see [section about `LPC subframes`](#subframelpc)) must be less than or equal to 12, i.e. the subframe type bits in the `subframe header` (see [section on `subframe header`](#subframeheader) may not be 101100-111111.
-  - The Rice partition order in a `Rice-coded residual section` (see [section on `Rice-coded residual section`](#residualcodingmethodpartitionedrice)) must be less than or equal to 8.
+  - The blocksize bits in the `FRAME_HEADER` (see [`FRAME_HEADER` section](#frameheader)) must be 0001-1110. The blocksize must be <= 16384; if the sample rate is <= 48000 Hz, the blocksize must be <= 4608.
```
retokromer: Do we really need three
```
`FRAME_HEADER` (see [`FRAME_HEADER` section](#frameheader)
```
in following lines? It makes the text completely unreadable.

More generally, do we need
```
`SOMETHING` (see [`SOMETHING` section](#something)
```
?

privatezero: The multiple references to FRAME_HEADER does seem excessive, but I left it in that format as that is what is present in the original spec.

The overly wordy format of `SOMETHING` (see [`SOMETHING` section](#something) is so that it will create an anchor that is intelligible both in the md and in the rfc created by the make file.

This way, after the make file is run it will look like `SOMETHING` (see Section 10.19).

retokromer: OK, let live it during the conformance to the «original».

dericed: If the term FRAME_HEADER is used before it is defined, then it should include a forward reference. For additional uses it doesn't necessarily need a reference every time it's used.

privatezero: I can dig it! Will adjust accordingly!

privatezero: Removed multiple FRAME_HEADER references as well as redundant links in the Architecture section. (Were referring to already referenced section and not specific term).

privatezero:
@retokromer I resolved the conflicts for this branch - does it still look ok to merge?

Issue #40, opened 2017-05-23
add RFC2119 section and implement the terminology


dericed:
An RFC2119 section SHOULD be added just after the introductory section, such as 

```
# Notation and Conventions

The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [@!RFC2119].
```

All current uses of those keywords with the specification SHOULD be reviewed to ensure that the use the same meaning as RFC2119 defines, see https://www.ietf.org/rfc/rfc2119.txt. If so then the term SHOULD be changed to all caps; if not, the the term should be changed so as not to confuse meaning with RFC2119.

PR #48, opened 2017-06-02
Rfc2119

dericed:
Here are a few places where I thought a different rfc2119 term was
intended. Review carefully!

Intended to resolve https://github.com/privatezero/flac_markdown/issues/40


flac.md
```diff
@@ -105,27 +109,27 @@ Before the formal description of the stream, an overview might be helpful.
...
+  - The blocksize bits in the `FRAME_HEADER` (see [`FRAME_HEADER` section](#frameheader)) MUST be 0001-1110. The blocksize MUST be <= 16384; if the sample rate is <= 48000 Hz, the blocksize MUST be <= 4608.
+  - The sample rate bits in the `FRAME_HEADER` MUST be 0001-1110.
+  - The bits-per-sample bits in the `FRAME_HEADER` MUST be 001-111.
+  - If the sample rate is <= 48000 Hz, the filter order in `LPC subframes` (see [`SUBFRAME_LPC` section](#subframelpc)) MUST be less than or equal to 12, i.e. the subframe type bits in the `SUBFRAME_HEADER` (see [`SUBFRAME_HEADER` section](#subframeheader)) SHOULD NOT be 101100-111111.
```
retokromer: The values in the previous 4 lines should begin with `0b`, because they are binary.

flac.md
```diff
@@ -225,38 +229,38 @@ Data     | Description
...
 `u(1)`            | `1` if the CUESHEET corresponds to a Compact Disc, else `0`.
-`u(7+258\*8)`     | Reserved. All bits must be set to zero.
-`u(8)`            | The number of tracks. Must be at least 1 (because of the requisite lead-out track). For CD-DA, this number must be no more than 100 (99 regular tracks and one lead-out track).
-`CUESHEET_TRACK`+ | One or more tracks. A CUESHEET block is required to have a lead-out track; it is always the last track in the CUESHEET. For CD-DA, the lead-out track number must be 170 as specified by the Red Book, otherwise is must be 255.
+`u(7+258\*8)`     | Reserved. All bits MUST be set to zero.
+`u(8)`            | The number of tracks. Must be at least 1 (because of the requisite lead-out track). For CD-DA, this number MUST be no more than 100 (99 regular tracks and one lead-out track).
+`CUESHEET_TRACK`+ | One or more tracks. A CUESHEET block is REQUIRED to have a lead-out track; it is always the last track in the CUESHEET. For CD-DA, the lead-out track number MUST be 170 as specified by the Red Book, otherwise is MUST be 255.
```
privatezero: looks like there was a typo here in the original spec that has held over. Should be 'otherwise it MUST'

flac.md
dericed: out of scope with the intent of the PR

flac.md
dericed: fixed and rebased

dericed:
This PR is updated. Please re-review.

flac.md
retokromer: OK. It was just to avoid other conflicts. I will open an _ad hoc_ PR.

retokromer:
Indeed, I tried four times to fix the few missed `0b` by NN, since the very first correction of them, which has also been done twice, by NN and by me. BTW, in my opinion, e.g. `0b00-0b11` would be better than the current `0b00 - 0b11` (with spaces before and after the dash). If you agree, I will fix that. 

I don’t know why – nor when – the order of the section has been modified.

PR #51, opened 2017-06-02
internal link

retokromer:
should fix https://github.com/privatezero/flac_markdown/commit/dbd8ead82ccdb27ed917750dc8b7b2c8d7ab1f7a#commitcomment-22372658


retokromer:
Works fine locally… with the following warnings:
```
WARNING: artwork outdented 1 characters to avoid overrunning right margin around input line 585)
WARNING: so many <ttcol>s in <texttable> that some words need to be split near line 718
```

privatezero:
looks good to me!

PR #56, opened 2017-06-12
Add notes about subframe sample size

### LINKED ISSUE #53: Suggested improvements of the current specs
Posted on 2017-06-06 in [flac-dev]:

```
Hi all,

I'm jumping in on this thread to make a few remarks about the spec. I
implemented a FLAC decoder by only looking at the spec, and I have a few
notes that would have saved me a lot of time if the spec had mentioned
them. They are obvious in hindsight, of course.

* If the channel assignment includes a difference channel, then the
subframe for that channel has one extra bit per sample in order to
encode the difference.

* The number of bits per sample for a subframe, is the number of bits
per sample of the frame, minus the number of wasted bits per sample of
the subframe (and possibly plus one for a difference channel).

I hope this helps future implementers.

Kind regards,

Ruud van Asseldonk
```

ruuda:
The contents of this PR are the same as those of my patch submitted to flac-dev. Closes #53.

### DIFF
--- a/flac.md
+++ b/flac.md
@@ -403,6 +403,8 @@ Value   | Description
 0b110   | 24 bits per sample
 0b111   | reserved
 
+For subframes that encode a difference channel, the sample size is one bit larger than the sample size of the frame, in order to be able to encode the difference between extreme values.
+
 ### FRAME HEADER RESERVED2
 Value | Description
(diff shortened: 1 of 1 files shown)


privatezero:
Thank you much @ruuda!

PR #66, opened 2017-07-08
add where 4608 comes from

retokromer:


MarcusJohnson91:
Where DOES 4608 come from?

retokromer:
96 kHz at 24 bit are 4608 kbps

PR #74, opened 2018-11-11
Clarified the frame/sample encoding after speaking to the inventor, Josh Coalson to clarify it.

MarcusJohnson91:


privatezero:
Thanks!

Issue #84, opened 2020-09-12
METADATA_BLOCK_STREAMINFO MD5 signature field is underspecified


fweimer:
In the reference implementation, the MD5 signature appears to cover the raw, undecoded byte stream. For example, if encoding starts with a WAVE format file, the signature is computed over a stream of signed, 16-bit little-endian words. This is contrary to the rest of the specification, which is based on network byte order.

At this point, it is probably best not to specify a computation method for this field, and say that is implementation-defined and may be used for decoding consistency checks. This would also avoid mentioning MD5, which should make the security people happy (although of course MD5 is not used in a security-relevant way here).

ktmf01:
> At this point, it is probably best not to specify a computation method for this field, and say that is implementation-defined and may be used for decoding consistency checks.

If it would be implementation-defined it is of no value for decoding consistency checks, as one cannot differentiate between a decoding problem or implementation difference.

> This would also avoid mentioning MD5, which should make the security people happy (although of course MD5 is not used in a security-relevant way here).

FLAC also uses CRC for integrity checks, which are of course even less secure. Is it just because MD5 long ago used to have a place in cryptographic applications that it cannot be used for an integrity check in a non-executable file?

PR #100, opened 2021-10-04
Define the way the md5sum is calculated

ktmf01:
This should address issue #84 

Reference:
1. run `metaflac --list-md5sum somefile.flac`
2. run `flac -d --force-raw-format --endian=little --sign=signed somefile.flac -o somefile.raw`
3. run `md5sum somefile.raw`
4. Results should be identical

### DIFF
--- a/flac.md
+++ b/flac.md
@@ -191,7 +191,7 @@ Data     | Description
 
 FLAC specifies a minimum block size of 16 and a maximum block size of 65535, meaning the bit patterns corresponding to the numbers 0-15 in the minimum blocksize and maximum blocksize fields are invalid.
 
-The MD5 signature is made by performing an MD5 transformation on the samples of all channels interleaved, represented in a signed, little-endian form. This interleaving is on a per-sample basis, so for a stereo file this means first the first sample of the first channel, then the first sample of the second channel, then the second sample of the first channel etc. In case the bit depth is not a whole number of bytes, samples are left padded (right-justified) with zeros to a whole number of bytes before transforming.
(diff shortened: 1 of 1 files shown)


flac.md
```diff
@@ -191,7 +191,7 @@ Data     | Description
 
 FLAC specifies a minimum block size of 16 and a maximum block size of 65535, meaning the bit patterns corresponding to the numbers 0-15 in the minimum blocksize and maximum blocksize fields are invalid.
 
-The MD5 signature is made by performing an MD5 transformation on the samples of all channels interleaved, represented in a signed, little-endian form. This interleaving is on a per-sample basis, so for a stereo file this means first the first sample of the first channel, then the first sample of the second channel, then the second sample of the first channel etc.
+The MD5 signature is made by performing an MD5 transformation on the samples of all channels interleaved, represented in a signed, little-endian form. This interleaving is on a per-sample basis, so for a stereo file this means first the first sample of the first channel, then the first sample of the second channel, then the second sample of the first channel etc. In case the bit depth is not a whole number of bytes, samples are left padded (right-justified) with zeros to a whole number of bytes before transforming.
```
fweimer, line 194: Maybe this is clearer (assuming that it actually matches the reference implementation):

> In case the bit depth is not a whole number of bytes, additional zero bits are inserted at the most-significant position until the bit depth is a multiple of 8. This padding is performed for each channel independently, and before byte encoding.

ktmf01: Yes, it does read cleaner. However, what do you mean by byte encoding?

fweimer: I tried to express that the padding happens on the sample values, not the output bytes.

PR #108, opened 2021-10-18
Add a security section

### LINKED ISSUE #105: Add a "Security Considerations" section
The issues found in files from https://wiki.hydrogenaud.io/index.php?title=FLAC_decoder_testbench may be a good starting point on what needs to be looked at.

JeromeMartinez:
Fix #105.

### DIFF
--- a/flac.md
+++ b/flac.md
@@ -576,3 +576,12 @@ The number of samples (n) in the partition is determined as follows:
 - if the partition order is zero, n = frame's blocksize - predictor order
 - else if this is not the first partition of the subframe, n = (frame's blocksize / (2\^partition order))
 - else n = (frame's blocksize / (2\^partition order)) - predictor order
+
+# Security Considerations
+
+Like any other codec (such as [@?RFC6716]), FLAC should not be used with insecure ciphers or cipher modes that are vulnerable to known plaintext attacks. Some of the header bits as well as the padding are easily predictable.
+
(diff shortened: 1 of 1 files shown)


flac.md
```diff
@@ -576,3 +576,12 @@ The number of samples (n) in the partition is determined as follows:
...
+See the [FLAC decoder testbench](https://wiki.hydrogenaud.io/index.php?title=FLAC_decoder_testbench) for a non exhaustive list of FLAC files with extreme configurations which lead to crashes or reboots on some known implementations.
```
retokromer, line 585: I guess (I am not English speaker), it should be `non-exhaustive list` (with a dash).

flac.md
```diff
@@ -576,3 +576,12 @@ The number of samples (n) in the partition is determined as follows:
...
+
+Implementations of the FLAC codec need to take appropriate security considerations into account. Those related to denial of service are outlined in Section 2.1 of [@!RFC4732]. It is extremely important for the decoder to be robust against malicious payloads. Malicious payloads **MUST NOT** cause the decoder to overrun its allocated memory or to take an excessive amount of resources to decode. An overrun in allocated memory could lead to arbitrary code execution by an attacker. The same applies to the encoder, even though problems in encoders are typically rarer. Malicious audio streams **MUST NOT** cause the encoder to misbehave because this would allow an attacker to attack transcoding gateways. An example is allocating more memory than available especially with blocksizes of more than 10000 or with big METADATA_BLOCK_VORBIS_COMMENT blocks, or not allocating enough memory before copying data, which lead to execution of malicious code, crashes, freezes or reboots on some known implementations.
```
ktmf01, line 584:
```suggestion
Implementations of the FLAC codec need to take appropriate security considerations into account. Those related to denial of service are outlined in Section 2.1 of [@!RFC4732]. It is extremely important for the decoder to be robust against malicious payloads. Malicious payloads **MUST NOT** cause the decoder to overrun its allocated memory or to take an excessive amount of resources to decode. An overrun in allocated memory could lead to arbitrary code execution by an attacker. The same applies to the encoder, even though problems in encoders are typically rarer. Malicious audio streams **MUST NOT** cause the encoder to misbehave because this would allow an attacker to attack transcoding gateways. An example is allocating more memory than available especially with blocksizes of more than 10000 or with big metadata blocks, or not allocating enough memory before copying data, which lead to execution of malicious code, crashes, freezes or reboots on some known implementations.
```
I'd keep it more general, as problems are even more likely to occur with huge SEEKTABLE blocks, and can also happen with PICTURE and APPLICATION blocks, depending on whether or not the decoder decides to parse them.

PR #117, opened 2021-11-18
Almost complete rewrite

### LINKED ISSUE #113: Add caution note about different bit size for METADATA_BLOCK_HEADER (24) and METADATA_BLOCK_PICTURE (32) length
There are software out there that ends up writing a truncated length in METADATA_BLOCK_PICTURE. ffmpeg used to write truncated size (now it fails instead) and now also has a workaround for decoding truncated files (https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/flac_picture.c#L129).

ktmf01:
A few weeks ago I started on a full rewrite of the specification in case copyright of the original specification would become a problem. As far as I know, nothing mentioned in the current document is missing from this rewrite, but also nothing has been copied that has been written be anyone not involved in this working group.

If the copyright problems are resolved, parts of this rewrite can of course be cherry-picked. Specifically, I was thinking about:

- the vorbis comment subsection, which is much more detailed than the current one
- the subframes subsection, which is mostly two of the pull requests that I just closed
- the picture metadata block subsection, which is now more specific

Moreover, the document has a different structure and much less tables. I cannot say all parts of this document benefit from the lack of tables per se (compare for example the specification of the streaminfo metadata block) so perhaps both approaches should be mixed.

It is probably too early to start grammar and spell checking, perhaps for now just commenting on the idea and the structure of the document?

### DIFF
--- a/flac.md
+++ b/flac.md
@@ -1,48 +1,22 @@
 # Introduction
 
-This is a detailed description of the FLAC format. There is also a companion document that describes [FLAC-to-Ogg mapping](https://xiph.org/flac/ogg_mapping.html).
+This document provides a detailed definition of the FLAC format. FLAC stands for Free Lossless Audio Codec: it is designed to reduce the amount of computer storage space needed to store digital audio signals without needing to remove information in doing so (i.e. lossless). FLAC is free in the sense that its specification is open, its reference implementation is open-source and it is not encumbered by any known patent.
 
-For a user-oriented overview, see [About the FLAC Format](https://xiph.org/flac/documentation_format_overview.html).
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
It seems that we don't have any copyright problem anymore, but IMO it is worth it to keep most of this PR. It will need a big review though.

Having a mix, e.g. by keeping tables for enums, may be better.

ktmf01:
One complication here is that, in order for the rewrite not to be derivative, the structure of the two documents is in some places quite different. The stream specification itself is mostly the same, because it follows the order of a FLAC stream, but parts like subset definition, format overview, explanation of channel decorrelation do not have direct counterparts in the other document or have a different 'depth' (section instead of subsubsection or vice versa)

ktmf01:
As far as I can see, everything that could be used directly has been merged in other pull requests. This rewrite can still be viewed after closing, so I'm closing this PR.

PR #120, opened 2021-11-25
Add appendix with example file decoding 'by hand'

ktmf01:
In a previous PR (and also in the rewrite), I added an example to the coded residual section which felt out of place. Such an example didn't seem to fit in a formal specification, so the idea came to write an appendix with more thorough examples.

These examples can provide a more hands-on way to understand the FLAC specification for readers that need it. Also, it can be used by people proofreading this specification to do cross-referencing, as it is 'redundant'.

If this turns out to be a welcome addition, I would like to add more examples to this appendix in the future, specifically:
- Example with a frame with a linear predictor (LPC)
- Example involving all other types of metadata blocks
- Example with variable blocksize frame
- Example with a bit depth that is not a whole number of bytes, specifically to explain how MD5summing works in that case

Please provide feedback


rfc_backmatter.md
```diff
@@ -1,2 +1,429 @@
...
+0xd6+0 | 16 bit | 0xc40b          | 16-bit unencoded sample
+0xd8+0 | 16 bit | 0xc14a          | 16-bit unencoded sample
+
+The second subframe starts at 0xda+0 and is broken down in the following table
+
+Start  | Length | Contents          | Description
+:------|:-------|:------------------|:-----------------
+0xda+0 | 1 bit  | 0b0               | mandatory 0 bit
+0xda+1 | 6 bit  | 0b000001          | verbatim subframe
+0xda+7 | 1 bit  | 0b0               | no wasted bits present
```
ktmf01, line 408:
```suggestion
0xda+7 | 1 bit  | 0b1               | wasted bits present
```
Correct copy-paste error

rfc_backmatter.md
```diff
@@ -1,2 +1,575 @@
 
 {backmatter}
+
+# Examples
+
+This appendix contains short example FLAC files and short parts of FLAC files which are decoded step by step. These examples provide a more engaging way to understand the FLAC format than the formal specification. The text explaining these examples assumes the reader has at least cursory read the specification and that the reader refers to the specification for explanation of the terminology used. These examples mostly focus on the lay-out of several metadata blocks and subframe types and the implications of certain aspects (for example wasted bits and stereo decorrelation) on this lay-out.
```
mcr, line 6:
```suggestion
This informational appendix contains short example FLAC files and short parts of FLAC files which are decoded step by step. These examples provide a more engaging way to understand the FLAC format than the formal specification. The text explaining these examples assumes the reader has at least cursory read the specification and that the reader refers to the specification for explanation of the terminology used. These examples mostly focus on the lay-out of several metadata blocks and subframe types and the implications of certain aspects (for example wasted bits and stereo decorrelation) on this lay-out.
```

ktmf01:
I just added commit da39c32 to this PR involving a change for the same cause as #129. However, instead of removing, the UTF-8 characters are changed (as this example really needs UTF-8) so it renders clearer when fed through mmark and xml2rfc. See commit message for details

rfc_backmatter.md
```diff
@@ -223,85 +225,85 @@ Start  | Length  | Contents           | Description
 0x44+0 | 4 byte  | 0x20000000         | Vendor string length 32 byte
 0x48+0 | 32 byte | (...)              | Vendor string
 0x68+0 | 4 byte  | 0x01000000         | Number of fields 1
-0x6c+0 | 4 byte  | 0x0d000000         | Field length 13 byte
-0x70+0 | 13 byte | (...)              | Field contents
+0x6c+0 | 4 byte  | 0x18000000         | Field length 24 byte
+0x70+0 | 24 byte | (...)              | Field contents
 
-The vendor string is reference libFLAC 1.3.3 20190804, the field contents of the only field is title=Québec. The vorbis comment field is 13 bytes but only 12 characters in size, because it contains one character needing 2 bytes to represent.
+The vendor string is reference libFLAC 1.3.3 20190804, the field contents of the only field is TITLE=Щелкунчик. The vorbis comment field is 24 bytes but only 15 characters in size, because it contains 9 character needing 2 bytes to represent.
```
retokromer, line 231: `TITLE= Матрёшка` would be funnier (sorry, I could not resist ;-)

ktmf01: Not a bad idea actually. I was looking for short words in scripts other than latin that would not be considered offensive in any way, and that was rather hard as I don't speak or read any Russian, Greek, Chinese etc.  Матрёшка is one character less then Щелкунчик so it is probably a better fit, as all those UTF-8 code points still make it look cluttered. Maybe you know something even shorter in Hebrew?

Now I think of it: If I shorten the vendor string, I don't have to change the whole example even with a longer vorbis comment field, so less room for error.

retokromer: Short, well known and all but offensive would be: שלום

ktmf01: Thanks, I'll add that in and force-push this PR branch to amend the last commit.

rfc_backmatter.md
```diff
@@ -1,2 +1,575 @@
...
+
+The vendor string is reference libFLAC 1.3.3 20190804, the field contents of the only field is TITLE=שלום. The vorbis comment field is 14 bytes but only 10 characters in size, because it contains four 2-byte characters.
+
+### Padding
+
+The last metadata block is a (very short) padding block.
+
+Start  | Length  | Contents           | Description
+:------|:--------|:-------------------|:-----------------
+0x7e+0 | 1 bit   | 0b0                | Last metadata block
```
ktmf01, line 237: Another copy-paste error
```suggestion
0x7e+0 | 1 bit   | 0b1                | Last metadata block
```

rfc_backmatter.md
```diff
@@ -1,2 +1,575 @@
...
+0x0a+0 | 2 byte  | 0x1000             | Max. blocksize 4096
+0x0c+0 | 3 byte  | 0x00000f           | Min. frame size 15 byte
+0x0f+0 | 3 byte  | 0x00000f           | Max. frame size 15 byte
+0x12+0 | 20 bit  | 0x0ac4, 0b0100     | Sample rate 44100 Hertz
+0x14+4 | 3 bit   | 0b001              | 2 channels
+0x14+7 | 5 bit   | 0b01111            | Sample bit depth 16
+0x15+4 | 36 bit  | 0b0000, 0x00000001 | Total no. of samples 1
+0x1a   | 16 byte | (...)              | MD5 signature
+
+The minimum and maximum blocksize are both 4096. This was apparently the blocksize the encoder was intending to use for this audio, but as only 1 interchannel sample was provided, no frames with size 4096 are actually present in this file. This is because even in fixed blocksize streams, the size of the last frame can be smaller.
```
ktmf01, line 71:
```suggestion
The minimum and maximum blocksize are both 4096. This was apparently the blocksize the encoder planned to use, but as only 1 interchannel sample was provided, no frames with 4096 samples are actually present in this file.
```

rfc_backmatter.md
```diff
@@ -1,2 +1,575 @@
...
+0x3e+3 | 5 bit  | 0b01000         | Residual 8
+0x3f+0 | 5 bit  | 0b00110         | Residual 6
+0x3f+5 | 4 bit  | 0b0010          | Rice parameter 2
+0x40+1 | 22 bit | (...)           | Residual partition 3
+0x42+7 | 4 bit  | 0b0001          | Rice parameter 1
+0x43+3 | 23 bit | (...)           | Residual partition 4
+
+The frame ends with 6 padding bits and a 2 byte frame CRC
+
+To decode this subframe, 21 predictions have to calculated and added to their corresponding residuals. This is a sequential process: as each prediction uses previous samples, it is not possible to start this decoding halfway a subframe or decode a subframe with parallel threads.
```
ktmf01, line 540:
```suggestion
To decode this subframe, 21 predictions have to be calculated and added to their corresponding residuals. This is a sequential process: as each prediction uses previous samples, it is not possible to start this decoding halfway a subframe or decode a subframe with parallel threads.
```

rfc_backmatter.md
```diff
@@ -1,2 +1,575 @@
...
+
+The second subframe starts at 0x34, it is broken down in the following table.
+
+Start  | Length | Contents        | Description
+:------|:-------|:----------------|:-----------------
+0x34+0 | 1 bit  | 0b0             | mandatory 0 bit
+0x34+1 | 6 bit  | 0b000001        | verbatim subframe
+0x34+7 | 1 bit  | 0b1             | wasted bits present
+0x35+0 | 4 bit  | 0b0001          | 4 wasted bits
+0x35+4 | 14 bit | 0b0010, 0x8b    | 12-bit unencoded sample
```
ktmf01, line 119:
```suggestion
0x35+4 | 12 bit | 0b0010, 0x8b    | 12-bit unencoded sample
```

retokromer: good catch!

rfc_backmatter.md
```diff
@@ -50,3 +50,575 @@ Taking the last example where 31 bits were needed for the prediction, the requir
...
+0x31+0 | 1 bit  | 0b0             | mandatory 0 bit
+0x31+1 | 6 bit  | 0b000001        | verbatim subframe
+0x31+7 | 1 bit  | 0b1             | wasted bits present
+0x32+0 | 2 bit  | 0b01            | 2 wasted bits
+0x32+2 | 14 bit | 0b011000, 0xfd  | 14-bit unencoded sample
+
+
+As the wasted bits flag is 1 in this subframe, an unary coded number follows. Starting at 0x32, we see 0b01, which unary codes for 1, meaning we have 2 wasted bits in this subframe.
+
+As this is a verbatim subframe, the subframe only contains unencoded sample values. With a blocksize of 1, it contains only a single sample. The bit depth of the audio is 16 bit, but as the subframe header signals 2 wasted bits, only 14 bits are stored. As no stereo decorrelation is used, a bit depth increase for the side channel is not applicable. So, the next 14 bit (starting at position 0x32+2) contain the unencoded sample coded big-endian, signed two’s complement. The value reads 0b011000 11111101, or 6397. This value needs to be shifted left by 2 bits, to account for the wasted bits. The value is then 0b011000 11111101 00, or 25588.
```
ktmf01, line 161:
```suggestion
As this is a verbatim subframe, the subframe only contains unencoded sample values. With a blocksize of 1, it contains only a single sample. The bit depth of the audio is 16 bit, but as the subframe header signals 2 wasted bits, only 14 bits are stored. As no stereo decorrelation is used, a bit depth increase for the side channel is not applicable. So, the next 14 bit (starting at position 0x32+2) contain the unencoded sample coded big-endian, signed two's complement. The value reads 0b011000 11111101, or 6397. This value needs to be shifted left by 2 bits, to account for the wasted bits. The value is then 0b011000 11111101 00, or 25588.
```

rfc_backmatter.md
```diff
@@ -50,3 +50,575 @@ Taking the last example where 31 bits were needed for the prediction, the requir
...
+:------|:--------|:-------------------|:-----------------
+0x04+0 | 1 bit   | 0b0                | Not the last metadata block
+0x08+0 | 2 byte  | 0x0010             | Min. blocksize 16
+0x0a+0 | 2 byte  | 0x0010             | Max. blocksize 16
+0x0c+0 | 3 byte  | 0x000017           | Min. frame size 23 byte
+0x0f+0 | 3 byte  | 0x000044           | Max. frame size 68 byte
+0x15+4 | 36 bit  | 0b0000, 0x00000013 | Total no. of samples 19
+0x1a   | 16 byte | (...)              | MD5 signature
+
+This time, the minimum and maximum blocksizes are reflected in the file: there is one block of 16 samples, but the last block (which has 3 samples) is excluded from this number. The MD5 signature is d5b0564975e98b8d8b930422757b8103, this will be verified at the end of this example.
```
ktmf01, line 251:
```suggestion
This time, the minimum and maximum blocksizes are reflected in the file: there is one block of 16 samples, but the last block (which has 3 samples) is excluded from this number. The MD5 signature is 0xd5b0 5649 75e9 8b8d 8b93 0422 757b 8103, this will be verified at the end of this example.
```

rfc_backmatter.md
```diff
@@ -50,3 +50,575 @@ Taking the last example where 31 bits were needed for the prediction, the requir
...
+0xa6+0 | 1 bit  | 0b1             | Quotient 0
+0xa6+1 | 11 bit | 0b01001110111   | Remainder 631
+0xa7+4 | 1 bit  | 0b1             | Quotient 0
+0xa7+5 | 11 bit | 0b01000100100   | Remainder 548
+0xa9+0 | 1 bit  | 0b1             | Quotient 0
+0xa9+1 | 11 bit | 0b01000010101   | Remainder 533
+0xaa+4 | 1 bit  | 0b1             | Quotient 0
+0xaa+5 | 11 bit | 0b00100001100   | Remainder 268
+
+At this point, the decoder should know it is done decoding the coded residual, as it received 16 samples, 1 warm-up sample and 15 residual samples. Each residual sample can be calculated from the quotient and remainder, and then undoing the zig-zag encoding. For example, the value of the first zig-zag encoded residual sample is 3 * 2^11 + 244 = 6388. As this is an even number, the zig-zag encoding is undone by dividing by 2, the residual sample value is 3194. This is done for all residual samples in the next table
```
ktmf01, line 358:
```suggestion
At this point, the decoder should know it is done decoding the coded residual, as it received 16 samples: 1 warm-up sample and 15 residual samples. Each residual sample can be calculated from the quotient and remainder, and undoing the zig-zag encoding. For example, the value of the first zig-zag encoded residual sample is 3 * 2^11 + 244 = 6388. As this is an even number, the zig-zag encoding is undone by dividing by 2, the residual sample value is 3194. This is done for all residual samples in the next table
```

PR #96, opened 2021-09-30
Replace tables under RESIDUAL with text and replace RESIDUAL with Coded residual

ktmf01:
This PR replaces all tables under RESIDUAL with a (quite long) explanation and an example. I've also replaced the term RESIDUAL with Coded residual.

Please provide me with some feedback as to whether the wording and style of this text is in line with what is expected of an RFC. I was thinking of moving away from the heap of tables and moving more towards the style in RFC 6716 (the Opus specification)

PR #122, opened 2021-12-12
Replace tables under RESIDUAL with text

ktmf01:
I tried reopening PR #96, but this was not possible as I force-pushed the branch. This force-push was to undo a change in the the first commit of this PR: changing RESIDUAL to Coded residual in the tables under SUBFRAME_FIXED and SUBFRAME_LPC. This was done to avoid merge conflicts with other PRs

Instead, I now rebuilt the whole branch. Added to the commits already in #96 are removing the example as it is probably better placed in an appendix like #120 and changing the wording slightly in the following instances:

> the partition order is [8, and the rest of the entropy block contains] 2^8 = 256 partitions -> 8 and the residual is split up into

> In case the coded residual of [the current subframe] -> a subframe

> containing the number of bits with which [the residual] is stored -> each residual sample

> the zigzag encoded residual [isbe called the] -> is called the


flac.md
```diff
@@ -544,61 +544,28 @@ Data      | Description
...
----------------:|:-----------
-0b0000 - 0b1110 | Exp-golomb parameter.
-0b1111          | Escape code, meaning the partition is in unencoded binary form using n bits per sample; n follows as a 5-bit number.
-
-### RESIDUAL_CODING_METHOD_PARTITIONED_EXP_GOLOMB2
-Data               | Description
-:------------------|:-----------
-`u(4)`             | Partition order.
-`EXP-GOLOMB2_PARTITION`+ | There will be 2\^order partitions.
+The partition order MUST be so that the blocksize is evenly divisible by the number of partitions. This means for example that for all uneven blocksizes, only partition order 0 is allowed.  The partition order also MUST be so that the (blocksize >> partition order) is larger than the predictor order. This means for example that with a blocksize of 4096 and a predictor order of 4, partition order cannot be larger than 9.
```
ktmf01, line 560:
```suggestion
The partition order MUST be so that the blocksize is evenly divisible by the number of partitions. This means for example that for all odd blocksizes, only partition order 0 is allowed.  The partition order also MUST be so that the (blocksize >> partition order) is larger than the predictor order. This means for example that with a blocksize of 4096 and a predictor order of 4, partition order cannot be larger than 9.
```

flac.md
```diff
@@ -544,61 +544,28 @@ Data      | Description
...
+In case the coded residual of a subframe is one with a 4-bit Rice parameter (see table at the start of this section), the first 4 bits of each partition are either a rice parameter or an escape code. These 4 bits indicate an escape code if they are 0b1111, otherwise they contain the rice parameter as an unsigned number. In case the coded residual of the current subframe is one with a 5-bit Rice parameter, the first 5 bits indicate an escape code if they are 0b11111, otherwise they contain the rice parameter as an unsigned number as well.
 
-#### EXP-GOLOMB2 PARTITION ENCODING PARAMETER
-Value             | Description
------------------:|:-----------
-0b00000 - 0b11110 | Exp-golomb parameter.
-0b11111           | Escape code, meaning the partition is in unencoded binary form using n bits per sample; n follows as a 5-bit number.
+In case an escape code was used, the partition does not contain a variable-length rice coded residual, but a fixed-length unencoded residual. Directly following the escape code are 5 bits containing the number of bits with which each residual sample is stored, as an unsigned number.
```
ktmf01, line 564:
```suggestion
In case an escape code was used, the partition does not contain a variable-length rice coded residual, but a fixed-length unencoded residual, with each residual sample signed two's complement. Directly following the escape code are 5 bits containing the number of bits with which each residual sample is stored, as an unsigned number.
```

PR #123, opened 2021-12-26
Incorrect subframe for decoding left subframe

MulattoKid:
When decoding the left subframe in a Right-Side encoded frame, the left subframe is decoded by adding the side subframe to the right subframe, not the left subframe (the one to decode).

### DIFF
--- a/flac.md
+++ b/flac.md
@@ -88,7 +88,7 @@ In many audio files, channels are correlated. The FLAC format can exploit this c
 
 - **Right-side**. The right subblock is coded and the left and right subblock are used to code a side subframe. Note that the actual coded subframe order is side-right. The side subframe is constructed in the same way as for mid-side. To decode, the left subblock is restored by adding the samples in the side subframe to the corresponding samples in the right subframe.
 
-The side channel needs one extra bit of bit depth as the subtraction can produce sample values twice as large as the maximum possible in any given bit depth. The mid channel in mid-side stereo does not need one extra bit, as it is shifted left one bit. The left shift of the mid channel does not lead to non-lossless behavior, because an uneven sample in the mid subframe must always be accompanied by a corresponding uneven sample in the side subframe, which means the lost least significant bit can be restored by taking it from the sample in the side subframe.
+The side channel needs one extra bit of bit depth as the subtraction can produce sample values twice as large as the maximum possible in any given bit depth. The mid channel in mid-side stereo does not need one extra bit, as it is shifted right one bit. The right shift of the mid channel does not lead to non-lossless behavior, because an uneven sample in the mid subframe must always be accompanied by a corresponding uneven sample in the side subframe, which means the lost least significant bit can be restored by taking it from the sample in the side subframe.
 
 # Prediction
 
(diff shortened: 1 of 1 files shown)


flac.md
```diff
@@ -86,7 +86,7 @@ In many audio files, channels are correlated. The FLAC format can exploit this c
 
 - **Left-side**. The left subblock is coded and the left and right subblock are used to code a side subframe. The side subframe is constructed in the same way as for mid-side. To decode, the right subblock is restored by subtracting the samples in the side subframe from the corresponding samples the left subframe.
 
-- **Right-side**. The right subblock is coded and the left and right subblock are used to code a side subframe. Note that the actual coded subframe order is side-right. The side subframe is constructed in the same way as for mid-side. To decode, the left subblock is restored by adding the samples in the side subframe to the corresponding samples in the left subframe.
+- **Right-side**. The right subblock is coded and the left and right subblock are used to code a side subframe. Note that the actual coded subframe order is side-right. The side subframe is constructed in the same way as for mid-side. To decode, the left subblock is restored by adding the samples in the side subframe to the corresponding samples in the right subframe.
```
ktmf01, line 89:
```suggestion
- **Right-side**. The right subblock is coded and the left and right subblock are used to code a side subframe. Note that the actual coded subframe order is side-right. The side subframe is constructed in the same way as for mid-side. To decode, the right subblock is restored by adding the samples in the side subframe to the corresponding samples in the right subframe.
```

Here's another one.

flac.md
```diff
@@ -86,7 +86,7 @@ In many audio files, channels are correlated. The FLAC format can exploit this c
...
 
 The side channel needs one extra bit of bit depth as the subtraction can produce sample values twice as large as the maximum possible in any given bit depth. The mid channel in mid-side stereo does not need one extra bit, as it is shifted left one bit. The left shift of the mid channel does not lead to non-lossless behavior, because an uneven sample in the mid subframe must always be accompanied by a corresponding uneven sample in the side subframe, which means the lost least significant bit can be restored by taking it from the sample in the side subframe.
```
ktmf01, line 91:
```suggestion
The side channel needs one extra bit of bit depth as the subtraction can produce sample values twice as large as the maximum possible in any given bit depth. The mid channel in mid-side stereo does not need one extra bit, as it is shifted right one bit. The right shift of the mid channel does not lead to non-lossless behavior, because an uneven sample in the mid subframe must always be accompanied by a corresponding uneven sample in the side subframe, which means the lost least significant bit can be restored by taking it from the sample in the side subframe.

Note that the right shift cannot be replaced by dividing by two, because integer division usually results in rounding towards zero, while a right shift of a two complements signed integer results in rounding towards negative infinity.
```

ktmf01 (commented):
Thanks a lot for pointing out this mistake of mine. I just spotted a few more, maybe you can add them to this PR

retokromer:
Good caches, @MulattoKid @ktmf01, thank you!

flac.md
MulattoKid: Added.

MulattoKid:
Added @ktmf01 find.

retokromer (approved):

PR #132, opened 2022-02-14
Clean up Makefile

ktmf01:
Remove unused variables, remove warning on minimum versions (which is most likely no longer accurate), unduplicate mmark command, unduplicate names, prepare for release 3


retokromer (approved):

ktmf01:
The replacement of the seriesInfo -> value was copied from the matroska-specification Makefile

retokromer (approved):
Even more elegant, thank you!

PR #135, opened 2022-02-18
Change status to standard

ktmf01:
According to the charter, the FLAC spec should be a proposed standard, but the past two drafts have been released as 'informational'.

Is this observation correct? Please comment


JeromeMartinez:
This is "informational" because we document the past and we don't create something new based on agreements here.

JeromeMartinez:
and after that, if we go to a new version, it would be a standard.
At least it is what I understand based on what we do on FFV1 (v0-v3 are informational, and v4 would be a standard).

ktmf01:
I guess the charter is out of date then? [here](https://datatracker.ietf.org/wg/cellar/charter/) and [here](https://datatracker.ietf.org/doc/charter-ietf-cellar/)

> Deliverables:
> - Informational specification for Matroska container format versions 1, 2 and 3 to IESG for publication
> - Standards Track specification for Matroska container format version 4 to IESG for publication
> - Informational specification for FFV1 video codec versions 0, 1 and 3 to IESG for publication
> - Standards Track specification for FFV1 video codec version 4 to IESG for publication
> - Standards Track specification for FLAC audio codec to IESG for publication

mcr:
The status of FLAC could be open to discussion.
Let me ask the question this way: can we make changes to this document which would render existing implementations invalid?
If we can not, then the IETF does not (yet) have change control of the specification, and it should be informational, because it documents existing practice.

ktmf01:
I can think of many reasons why that wouldn't be desirable, but I cannot think of any (legal) reason why this document couldn't introduce breaking changes?

This quickly turns into a philosophical issue: if no legal barriers exist (and I think that is the case here), would 'acceptance by software vendors and the general public' be considered a reason a breaking change cannot be introduced? If that is the case, it very hard to tell in advance what the uptake will be. In other words: I don't know what the IETF considers barriers to having change control, so I can't answer the question.

Anyway, I just noticed that with an overhaul of the formatting of rfc_frontmatter.md in commit 697ff26, the status was changed from standard track to informational. There was no mention of this change in the commit message, in the meeting minutes or on the mailing list, so I thought it could have been a mistake. That's all.

mcr:
> I can think of many reasons why that wouldn't be desirable, but I cannot think of any (legal) reason why this document couldn't introduce breaking changes?

I don't know what legal means here.
IETF specifications are voluntary standards, and on their own, do not have any legal mandates.
So, let's drop that part.

IETF standards track internet-drafts are not stable, and implementers can and should expect breaking changes until they reach WGLC and RFC publication.  That's what it means for the IETF to have change control.  Once published, they are immutable, but of course, can be obsoleted or updated with a new document.

If this document is intended to document what is implemented now, and can not change any (or many) of the bits on the wire without breaking current implementations, then that implies that we can't change things.  Such documents are generally published as Informational document.  One of my first RFCs, 4322, was such a document. 

It is up to the WG to decide what kind of document we should have here.

ktmf01:
As discussed in the conversation here https://mailarchive.ietf.org/arch/msg/cellar/JrZXAa9YvCtD33rgIY9ADD45eMs/ this PR is being reopened

JeromeMartinez (approved):

retokromer (approved):

PR #136, opened 2022-02-18
Rewrite prediction section

### LINKED ISSUE #114: "run-length encoded" for constant subframe might be confusing
Currently says "The signal is run-length encoded and added to the stream." i remember reading it thinking it might be some generic run length encoding but it's just one constant with length of the block size as i understand it?

ktmf01:
This fixes #114. The section is rewritten to have less emphasis on the encoding aspect adds a basic explanation of predictor use and function.


flac.md
```diff
@@ -89,15 +89,17 @@ The side channel needs one extra bit of bit depth as the subtraction can produce
...
+1. **Linear prediction**. Samples are predicted using past samples and a set of predictor coefficients, the error of this prediction is processed by the residual coder. Compared to a fixed predictor, using a generic linear predictor adds overhead as predictor coefficients need to be stored. Therefore, this method of prediction is best suited for predicting more complex waveforms, where the added overhead is offset by space savings in the residual coding stage resulting from more accurate prediction. A linear predictor in FLAC has two parameters besides the predictor coefficients and the predictor order: the number of bits with which each coefficient is stored (the coefficient precision) and a prediction right shift. A prediction is formed by taking the sum of multiplying each predictor coefficient with the corresponding past sample, and dividing that sum by applying the specified right shift. For more information see the [section on the linear predictor subframe](#linear-predictor-subframe)
```
ktmf01, line 100:
```suggestion
1. **Linear predictor**. Samples are predicted using past samples and a set of predictor coefficients, the error of this prediction is processed by the residual coder. Compared to a fixed predictor, using a generic linear predictor adds overhead as predictor coefficients need to be stored. Therefore, this method of prediction is best suited for predicting more complex waveforms, where the added overhead is offset by space savings in the residual coding stage resulting from more accurate prediction. A linear predictor in FLAC has two parameters besides the predictor coefficients and the predictor order: the number of bits with which each coefficient is stored (the coefficient precision) and a prediction right shift. A prediction is formed by taking the sum of multiplying each predictor coefficient with the corresponding past sample, and dividing that sum by applying the specified right shift. For more information see the [section on the linear predictor subframe](#linear-predictor-subframe)
```

retokromer: +1

Issue #116, opened 2021-11-12
Including a list of open-source implementations?


ktmf01:
Would it be of any benefit to include a list of (high-quality) open-source decoder implementations? Or would this be something too volatile? I'd think of something like this:

- libFLAC (C and C++, BSD-like license)
- ffmpeg (C, GNU LGPL)
- Firefox (C, MPL)
- dr_libs (C. Public Domain/MIT No Attribution)
- Claxon (Rust, Apache 2.0)
- jFLAC (Java, LGPL)

retokromer:
I would include at least a reference implementation which, I presume, is `libFLAC`.

SpencerDawkins:
For @ktmf01 - If you're thinking the list is valuable, but volatile, and the FLAC community is willing to maintain the list, one option is to pick a (stable) place to put the list you're maintaining, and include a pointer to that list in the FLAC specification.

For @retokromer - if there is a stable place to look for libFLAC, you could definitely put a pointer to the reference implementation in the FLAC specification. 

Do those suggestions make sense?

mcr:
One stable place is the github wiki (here).
Another stable place is the new IETF wiki (not yet deployed. Old one was aweful).

ValZapod:
> * ffmpeg

We have some crazy bugs in it. But yes it is more effective than even reference with AVX patches. https://trac.ffmpeg.org/query?status=!closed&keywords=~flac

retokromer:
I actually use the FFmpeg implementation for teaching, but in production we use the libFLAC library inside FFmpeg.

wader:
Hello, i've been working on a tool for inspecting and querying binary files for a while that has quite good FLAC support that might be useful to other, it can be found here https://github.com/wader/fq.
My main uses of it currently is to debug broken media file, decoders, demuxers etc but also do various statistics like packing overhead, extract data for potting and so on.

JeromeMartinez:
> Hello, i've been working on a tool for inspecting and querying binary files for a while that has quite good FLAC support that might be useful to others, it can be found here https://github.com/wader/fq.

Wow, it is huge work. And an alternative to [MediaTrace](https://mediaarea.net/MediaTrace).
(I see that you listed it in related projects :-p)

Test files and a description about what's wrong with them could be added to https://github.com/ietf-wg-cellar/flac-test-files .

***

About a list, it may be useful to have it and volatility is fine in my opinion, it is a base for users. As well as a list on debugging tools.

[...]

PR #143, opened 2022-03-18
Add implementation status section, as per RFC 7942

ktmf01:


ktmf01:
This was added [as discussed on the mailing list](https://mailarchive.ietf.org/arch/browse/cellar/) and it fixes #116

retokromer (approved):

mcr (approved):

PR #147, opened 2022-04-27
Bump version to 04 and change NL to NLD

ktmf01:
The Internet-Draft submission complained about not recognizing NLD while xml2rfc does. Therefore this commit changes it to NL


retokromer (approved):
“Funny.”

Issue #146, opened 2022-03-29
Limit subframe bit-per-sample to 32


ktmf01:
Similar to issue #145, all FLAC decoders I know of handle subframe samples of at most 32-bit signed int. However, the FLAC specification does not forbid the use of stereo decorrelation together with 32-bit samples. When those are used together, a 33-bit per sample subframe results.

Perhaps forbidding the use of stereo decorrelation together with 32-bit samples would be best.

ktmf01:
Experimentation has shown that allowing 33-bit subframes does result in non-negligible compression improvement. See [discussion on the flac-dev mailinglist](http://lists.xiph.org/pipermail/flac-dev/2022-April/006528.html). Therefore it seems unwise to add this limitation to the specification.

PR #148, opened 2022-04-27
Add limit to residual, explain how to limit and when to check

### LINKED ISSUE #145: Limit residual to 32 bit signed int
All FLAC decoders I know of handle residuals of at most 32-bit signed int. However, the FLAC specification does not limit the size of residuals, they are only limited by the math from which they follow. In theory, residuals could reach a size equal to a 53-bit signed int: 32 + 1 + 15 + 5 - 0 = 53 bit (max bit depth + extra bit for side channel + max lpc coeff precision + log2(max_order) - min lpc shift)

Allowing these large residuals increases either decoding complexity (by requiring an extra data path to handle these residuals) or decreases decoder performance (by widening existing datapaths). Forbidding these large residuals slightly increases encoder complexity by adding a check to the created residuals.

I've [submitted a patch to ffmpeg to allow larger residuals](https://patchwork.ffmpeg.org/project/ffmpeg/patch/20220327133933.43039-1-mvanb1@gmail.com/) but on second thought limiting it is perhaps a better option. The issue is also being discussed on [the flac-dev mailinglist here](http://lists.xiph.org/pipermail/flac-dev/2022-March/006517.html) and [has been raised on libFLAC github](https://github.com/xiph/flac/issues/306)

ktmf01:
This fixes #145. This could be regarded as a breaking change to the format, but as this is mostly a problem with coding 32-bit PCM, of which no non-experimental implementations yet exist, I assume this won't be breaking anything.


JeromeMartinez:
In http://lists.xiph.org/pipermail/flac-dev/2022-March/006526.html you say "limiting to "31-bit int" is needed for backward
compatibility with ffmpeg", I don't follow all, does this change permits that old FFmpeg can decode any "valid" file or is there still an issue with some extreme files?

In https://github.com/xiph/flac/issues/306 you indicated a 24-bit file which has a 35-bit int, do I understand well if I say that libFLAC may create such file but can not decode it? Or is it impossible with any public version of libFLAC encoder?

It would also be interesting to know what is the behavior of other (car?) players with such file as an argument for limiting the encoder.

My main concern is that all major library maintainers accept this limitation. libFLAC, FFmpeg?

JeromeMartinez:
In https://patchwork.ffmpeg.org/project/ffmpeg/patch/20220103202608.56364-1-mvanb1@gmail.com/ you indicate "Stereo decorrelation is disabled on 32 bit-per-sample, because a side channel would need 33 bit-per-sample", is it correlated or handled somewhere else?

ktmf01:
> In http://lists.xiph.org/pipermail/flac-dev/2022-March/006526.html you say "limiting to "31-bit int" is needed for backward compatibility with ffmpeg", I don't follow all, does this change permits that old FFmpeg can decode any "valid" file or is there still an issue with some extreme files?

I amended that statement in the next mail in the chain: http://lists.xiph.org/pipermail/flac-dev/2022-April/006527.html Limiting to "31-bit int" isn't enough to guarantee backward compatibility. It is possible to create 32-bit FLAC files that play with old versions of libFLAC (back to 1.2.1) and ffmpeg (back to May 2015) but with the additional constraint that predictions cannot exceed the range of a 32-bit int, fixed subframes cannot be used and stereo decorrelation must not be used. Putting all these constraints in the spec just because two decoders that were never meant to decode 32-bit FLACs can decode them anyway seems a bit much.

> In xiph/flac#306 you indicated a 24-bit file which has a 35-bit int, do I understand well if I say that libFLAC may create such file but can not decode it? Or is it impossible with any public version of libFLAC encoder?
> 
> It would also be interesting to know what is the behavior of other (car?) players with such file as an argument for limiting the encoder.

Currently the FLAC specification does not limit residual samples to any size, only implicitly. In theory, the maximum residual sample that can occur is with a 33 bit subframe, 15 bit predictor coefficients, 32 order predictor and a prediction right shift of 0. In that case, the maximum residual sample possible needs 33+14+log(2;32)-0+1 = 53 bit.

However, this is purely theoretical: because of the way libFLAC and ffmpeg implement encoding, the prediction right shift is generally only a bit smaller than the predictor coefficient precision. Usually, subframes with a predictor coefficient precision of 15 have a prediction right shift of between 11 and 14. For 25-bit subframes (sideframes of 24-bit PCM) this results in 25+14+log(2;32)-11+1 = 34 bit. This still isn't usually reached, because 32nd order predictors aren't used by default and not all predictor coefficients will be maxed out.

The reason I stumbled upon this problem is that I've been working on a different way to construct linear predictors, aiming at higher compression. With this method, which is able to find more redundancy, a 'cleaner' predictor can be found, and the difference between predictor coefficient precision and prediction right shift can be larger. The 35-bit residual mentioned here was created with that experimental compressor combined with some crafted input material.

It might be possible to craft 24-bit PCM input to get a current released libFLAC to output residual samples that exceed the range of a 32-bit int, but there will be integer overflow problems already in the encoder when that happens. So, even if decoders are amended to accept residual samples exceeding the range offered by a 32-bit int, files created with currently released version of libFLAC with problematic residuals don't properly decode, as the encoder has already mangled them. So, residuals are currently already limited, just not in a fashion that is properly decodable.

> My main concern is that all major library maintainers accept this limitation. libFLAC, FFmpeg?

Why do you think acceptance is a concern? Encoders and decoders of these libraries are already de-facto limited, properly implementing a limit seems to me less intrusive than writing new, wider datapaths.

> In https://patchwork.ffmpeg.org/project/ffmpeg/patch/20220103202608.56364-1-mvanb1@gmail.com/ you indicate "Stereo decorrelation is disabled on 32 bit-per-sample, because a side channel would need 33 bit-per-sample", is it correlated or handled somewhere else?

This is an additional constraint for backward compatibility with existing versions of libFLAC and ffmpeg. See issue #146 as well. At first I thought this additional constraint should be added to the specification, but it turns out limiting this would reduce compression by about 1.6%, which I think is too much. Limiting the residual to 32-bit int has a neglicable (<0.01%) impact on compression however. See http://lists.xiph.org/pipermail/flac-dev/2022-April/006528.html for details.

JeromeMartinez:
> The reason I stumbled upon this problem is that I've been working on a different way to construct linear predictors,

And it is good to catch the issue now than later! Pushing the limits of the format and see the issues then discussing about them it very good, thank you for your hard work.

>  Putting all these constraints in the spec just because two decoders that were never meant to decode 32-bit FLACs can decode them anyway seems a bit much.

My concern is to output with a new encoder a stream that an old version FFmpeg believes as it can decode it but it has issues so the bit-to-bit reversibility is broken without warnings. And I understand that with 32-bit input it will be often.
Will it be fine with your FFmpeg patch for 32-bit residuals?
How does older FFmpeg behave with such file?

I understand the wish to avoid this 31-bit limit, but IMO here we know that it is a problem with used decoders and we need to be careful and document the limitations if we don't limit the encoders output.

> > My main concern is that all major library maintainers accept this limitation. libFLAC, FFmpeg?

> Why do you think acceptance is a concern? Encoders and decoders of these libraries are already de-facto limited, properly implementing a limit seems to me less intrusive than writing new, wider datapaths.

Issue is that if a well known encoder, and especially a reference decoder, does not follow the spec, the spec may be considered as untrustworthy. What would happen if libFLAC maintainers decide to implement 32-bit input support with 64-bit residuals and predictions?
It is a corner case so I am fine to limit to 32-bit int there if we have:
- agreement from libFLAC maintainers that it is the way to go, preferably with a change in the current 24-bit input support for blocking the virtually impossible case of 33+ bit stuff
- agreement from FFmpeg maintainers that it is the way to go, preferably with a change in the current 24-bit and 32-bit input support for blocking the 33+ bit stuff
- correct support in FFmpeg of stereo decorrelation when 32-bit stuff is used

In summary: my understanding of the issue is not complete but I am afraid of different behaviors depending on the well known decoder used, and I am afraid that their maintainers don't agree and provide a lot of "non conforming" streams.

In practice, having a 32-bit patch accepted in libFLAC and FFmpeg with theses limitations in mind (and the same limitations) for both encoder and decoder would be great.
If old 2021 FFmpeg does not support all streams, it is a pity but not something blocking IMO, maybe an option in libFLAC and FFmpeg would be helpful there (and it would make the issue more known by users).

 I know, it is not an easy task and my skills for helping there are very limited :(.

ktmf01:
> > Putting all these constraints in the spec just because two decoders that were never meant to decode 32-bit FLACs can decode them anyway seems a bit much.
> 
> My concern is to output with a new encoder a stream that an old version FFmpeg believes as it can decode it but it has issues so the bit-to-bit reversibility is broken without warnings. And I understand that with 32-bit input it will be often. Will it be fine with your FFmpeg patch for 32-bit residuals? How does older FFmpeg behave with such file?
>
> I understand the wish to avoid this 31-bit limit, but IMO here we know that it is a problem with used decoders and we need to be careful and document the limitations if we don't limit the encoders output.

I see, I didn't think of it that way. I see two possible solutions
1) Add limitations to the spec that subframes must only use fixed predictors when `predictor_order + subframe_bps <= 32`, residuals must not exceed `31-bit int`, stereo decorrelation cannot be used on 32 bps streams and predictions cannot exceed the range of a 32-bit int
2) Assign one of the [currently reserved values](https://www.ietf.org/archive/id/draft-ietf-cellar-flac-03.html#name-bit-depth-bits) for bit-depth to 32-bit streams. That way, current decoders do not know how to handle such input.

The problem with the first solution is that it severely hampers compression (by about 2%) and these seemingly arbitrary limitations are ugly in a specification.

The problem with the second solution is that is doesn't work for streams with a bps between 25 and 31. I think this will affect much less people.

> > > My main concern is that all major library maintainers accept this limitation. libFLAC, FFmpeg?
> 
> > Why do you think acceptance is a concern? Encoders and decoders of these libraries are already de-facto limited, properly implementing a limit seems to me less intrusive than writing new, wider datapaths.
> 
> Issue is that if a well known encoder, and especially a reference decoder, does not follow the spec, the spec may be considered as untrustworthy.

Yes, that is a problem. I meant the question as in: why would they not accept this solution, as the alternative (encoding and decoding 64-bit residuals) would add a lot of complexity for very little gain. Especially augmenting the bitreader can give rise to (security related) bugs.

> It is a corner case so I am fine to limit to 32-bit int there if we have:
> 
> agreement from libFLAC maintainers that it is the way to go, preferably with a change in the current 24-bit input support for blocking the virtually impossible case of 33+ bit stuff

I've discussed in on the mailinglist and the only person that responded (who is a long-standing contributor to libFLAC) agreed with this. I've already got rough patches for libFLAC ready for implementing said block, both in encoder and decoder.

Considering libFLAC maintainers: it currently has none. The previous maintainer is otherwise occupied. I myself got write access to the libFLAC github repo a few weeks ago and am working on all open issues and PRs, which is why I have been less active on this repo recently. As I've already discussed this topic on the mailinglist, I don't think there will be much against implementing this limit.

>agreement from FFmpeg maintainers that it is the way to go, preferably with a change in the current 24-bit and 32-bit input support for blocking the 33+ bit stuff

I could sent in a patch in a few weeks time that covers this, and see what discussion results.

JeromeMartinez:
> Yes, that is a problem. I meant the question as in: why would they not accept this solution, as the alternative (encoding and decoding 64-bit residuals) would add a lot of complexity for very little gain. Especially augmenting the bitreader can give rise to (security related) bugs.

Another person may think that the "little gain" is worth it. So we need to be clear with major libs (libFLAC, FFmpeg, other?) that this limitation is accepted. Having the corresponding code in libFLAC and FFmpeg for their encoders would be a good step IMO, and we make it explicit in the spec.

> The problem with the first solution is that it severely hampers compression (by about 2%) and these seemingly arbitrary limitations are ugly in a specification.

The 2% are for 32-bit, right? IMO ugly spec is better than files not played in old (including players embedded in cars) players, issue there is that we don't start for nothing and FLAC is already widely spread for 16 and 24 bps.
I tested your crafted 24-bit file `Extreme residual.flac` with Firefox and VLC, it is difficult to listen in Firefox and awful in VLC. This kind of stream should be out of spec, not enough support. Could you provide a sample file which is max 32-bit but not decoded in FFmpeg? I am curious about how it is with Firefox (not relying on FFmpeg on Windows IIUC).

Maybe I say stupid stuff, I don't understand all, don't hesitate to correct me, but from hat I understand we could:
- everything <=24 bps (what is spread nowadays): we limit to what is supported in libFLAC/FFmpeg/Firefox/cars; does it means 31 or 32? I understand that 31 would be better there (for FFmpeg only? not cars?) and if it does not hurt compression it may be better for backward compatibility.
- >24 bps and <32 bps: not sure there... as 24 bps or as 32 bps?
- 32 bps (what will be actually used): specific bit depth value in sub frame header, and 32 (so 33 for stereo correlation?) everywhere.

> I don't think there will be much against me merging my own patches.

So fine :). But FFmpeg is also very important there.

***

After that, having a test suite for all discussed cases and test a bit everywhere is the next part, a long trip... But without that I am afraid to be able to say that FLAC is limited to 24-bit in practice :(.

ktmf01:
> The 2% are for 32-bit, right? IMO ugly spec is better than files not played in old (including players embedded in cars) players, issue there is that we don't start for nothing and FLAC is already widely spread for 16 and 24 bps.

Compared to the list [here](https://wiki.hydrogenaud.io/index.php?title=FLAC_decoder_testbench), I think not being able to play 32-bit files is the least worrying of the whole list of problems embedded players tend to have. However, it is reasonable to assume that most embedded players will gracefully fail when the streams are marked with a codepoint that is currently reserved.

> I tested your crafted 24-bit file `Extreme residual.flac` with Firefox and VLC, it is difficult to listen in Firefox and awful in VLC. This kind of stream should be out of spec, not enough support. Could you provide a sample file which is max 32-bit but not decoded in FFmpeg? I am curious about how it is with Firefox (not relying on FFmpeg on Windows IIUC).

Yes, that is file 63 of the PR here: https://github.com/ietf-wg-cellar/flac-test-files/pull/3 That plays fine in libFLAC but not in ffmpeg. This was also made with an experimental encoder.

> Maybe I say stupid stuff, I don't understand all, don't hesitate to correct me, but from hat I understand we could:
> 
>   everything <=24 bps (what is spread nowadays): we limit to what is supported in libFLAC/FFmpeg/Firefox/cars; does it means 31 or 32? I understand that 31 would be better there (for FFmpeg only? not cars?) and if it does not hurt compression it may be better for backward compatibility.

This works two ways: when limiting the residual to 31-bit ints, current ffmpeg can decode streams, so that improves compatibility. However, existing versions of libFLAC are currently (implicitly) limited at 32-bit ints, so they violate the spec. That means that placing a limit at 32 is more compatible from a libFLAC encoder standpoint. I still think this is only an issue with crafted files and encoders.

>  .  > 24 bps and <32 bps: not sure there... as 24 bps or as 32 bps?

Hard to say, can't think of any use for these bitdepths, so it's hard to say who uses it in what way.

> 32 bps (what will be actually used): specific bit depth value in sub frame header, and 32 (so 33 for stereo correlation?) everywhere.

I'd say also limit residual sample bit depth to 32 for side frames, otherwise the extra bitwriter/bitreader is necessary anyway

JeromeMartinez:
> Yes, that is file 63 of the PR here: https://github.com/ietf-wg-cellar/flac-test-files/pull/3 That plays fine in libFLAC but not in ffmpeg.

And stall in Firefox Windows (other files where playing a bit), not good for compatibility.

>  I still think this is only an issue with crafted files and encoders.

Additional reason for forbidding it, IMO, as they don't play well on software, this is not useful to permit it if it is not well supported and it does not provide any better compression.
But true, if a new FFmpeg supports this file, we may consider that it is just a matter of time and we'll provide this file in the test package for people willing to test their decoder.
2 possibilities to consider, I am fine with both.

> > 24 bps and <32 bps: not sure there... as 24 bps or as 32 bps?

> Hard to say, can't think of any use for these bitdepths, so it's hard to say who uses it in what way.

If we don't put a limit for <24 bps, there is no choice to do there as it is same everywhere, right?

> I'd say also limit residual sample bit depth to 32 for side frames, otherwise the extra bitwriter/bitreader is necessary anyway

I don't follow, [stereo decorrelation part](https://www.ietf.org/archive/id/draft-ietf-cellar-flac-03.html#appendix-A.2) says that it needs +1 bit, how is it handle for 32 bps streams in order to keep max 32-bit everywhere?

rfc_backmatter.md
```diff
@@ -44,9 +44,15 @@ For subframes with a linear predictor, calculation is a little more complicated.
...
+
+For the residual of a fixed predictor, the maximum size of a residual was already calculated in the previous section. However, for a linear predictor, the prediction is shifted right by a certain amount. The number of bits needed for the residual is the number of bits calculated in the previous section, reduced by the prediction right shift, increased by one bit to account for the subtraction of the prediction from the current sample on encoding.
+
+Taking the last example of the previous section, where 31 bits were needed for the prediction, the required data type size for the residual samples in case of a right shift of 10 bits would be 31 - 10 + 1 = 22 bits, which means it is not necessary to check whether the residuals fit a 32-bit signed integer.
+
+As another example, when encoding 32-bit PCM with fixed predictors, only the 0-order fixed predictor is guaranteed to have residuals that fit a 32-bit signed int, while the residual samples produced by all other fixed predictors need to be checked for this.
```
ktmf01, line 55:
```suggestion
As another example, when encoding 32-bit PCM with fixed predictors, all predictor orders must be checked. While the 0-order fixed predictor is guaranteed to have residuals that fit a 32-bit signed int, it might produce a residual being the most negative representable value of that 32-bit signed int.
```

ktmf01 (commented):
I've been experimenting with 32-bit FLACs, and found one possible problem: INT32_MIN, i.e. the negativemost possible value of a 32-bit signed int. There are a few problems associated with it: negating it isn't possible without resorting to a wide data type, most `abs` functions in C runtime libraries have undefined behavior when processing INT32_MIN. Also, both ffmpeg's decoding as well as libFLAC's encoding become a bit more complicated.

Please comment, both on the concept of disallowing usage of a INT32_MIN as well as the wording I used.

PR #149, opened 2022-04-28
Allow floats in vorbis comment

ktmf01:
The text could be interpreted such that floating-point numbers would not be allowed in vorbis comment fields. However,  ReplayGain does just that, and there is no reason not to allow that. This commit clarifies that vorbis comment fields are not restricted


flac.md
```diff
@@ -141,7 +141,7 @@ The FLAC format uses two forms of Rice coding, which only differ in the number o
...
 
-All numbers used in a FLAC bitstream MUST be integers; there are no floating-point representations. All numbers MUST be big-endian coded, except the length field used in Vorbis comments, which MUST be little-endian coded. All numbers MUST be unsigned except linear predictor coefficients, the linear prediction shift and numbers which directly represent samples, which MUST be signed. None of these restrictions apply to application metadata blocks.
+All numbers used in a FLAC bitstream MUST be integers; there are no floating-point representations. All numbers MUST be big-endian coded, except the field length used in vorbis comments, which MUST be little-endian coded. All numbers MUST be unsigned except linear predictor coefficients, the linear prediction shift and numbers which directly represent samples, which MUST be signed. None of these restrictions apply to application metadata blocks or to vorbis comment field contents.
```
retokromer, line 144: Shouldn’t the first letter of `Vorbis` be capitalised?

retokromer (approved):

Issue #112, opened 2021-10-20
Document mapping into containers


JeromeMartinez:
As [we did for FFV1](https://www.rfc-editor.org/rfc/rfc9043.html#section-4.3.3), we should document mapping into common containers in order to be clear about how to do so.

"fLaC", METADATA_BLOCK_STREAMINFO, METADATA_BLOCK in track header
FRAME in container "blocks".

And document the known issues e.g. METADATA_BLOCK_STREAMINFO info is sometimes (file "cut") not relevant.

ktmf01:
I can write something for ogg, perhaps the matroska-codecs RFC should be referenced, that's all well defined. I can't figure out the status of FLAC-in-MP4 though, it seems quite a mess with lots of bits and pieces. I have zero knowlegde of MP4, so I'd appreciate some help. Does anybody know who to contact on this matter?

There is a draft mapping: https://github.com/xiph/flac/blob/master/doc/isoflac.txt 
There is an experimental muxer in ffmpeg
There is apparently an issue with the draft mapping: https://github.com/xiph/flac/issues/38 and https://github.com/w3c/media-source/issues/188

I've read the issue reports multiple times, but I don't understand what is what, and what the actual problem is. Similarly, I'm at a loss about whether code-points can be already registered, or whether this should happen after the flac specification becomes an RFC.

phschafft:
Little note on mappings: The current documents are a bit unclear to me how the `SEEKPOINT`s are to be understood when in a container. Also given the problem of re-muxing which might change the offsets on the physical bitstream but not on the logical bitstream.

ktmf01:
I would argue seekpoints are not in any way usable for FLAC files embedded in a container. First because the container will provide its own seeking method, and second because as you say the offsets are changed on remuxing, especially when other streams like video are also placed in that container. The seektable is embedded in Ogg/Matroska/MP4 streams but it serves no purpose other than being handy on demuxing the FLAC stream back to a file (which I presume is not a use case encountered often).

Case in point: libFLAC only uses a seektable in a non-embedded FLAC stream. For Ogg FLAC streams, it ignores the seektable altogether.

Perhaps a note should be added to the spec that a seektable metadata block may be encapsulated, but that is serves no purpose at all? Maybe we should even recommend to drop the seektable on embedding?

phschafft:
I think clarification is needed. However I'm not sure about "that is serves no purpose at all":
[...]

PR #158, opened 2022-05-23
Add details on Ogg and Matroska mapping, refer for MP4

ktmf01:
MP4 needs to be added as well, but as noted in #112, I need some help with that.


flac.md
```diff
@@ -630,6 +630,28 @@ Decoding the coded residual thus involves selecting the right coding method, fin
...
 
+# Container mappings
+
+The FLAC format can be used without any container as the FLAC format already provides for a very thin transport layer. However, the functionalty of this transport is rather limited, and to be able to combine FLAC audio with video, it needs to be encapsulated by a more capable container. This presents a problem: the transport layer provided by the FLAC format mixes data that belongs to the encoded data (like blocksize and sample rate) with data that belongs to the transport (like checksum and timecode). The choice was made to encapsulate FLAC frames as they are, which means some data will be duplicated and potentially deviating between the FLAC frames and the encapsulating container.
+
+To differentiate encapsulated FLAC streams from non-encapsulated FLAC streams, the latter will be referred to as 'native FLAC'.
```
JeromeMartinez, line 637: 'native FLAC' --> 'raw FLAC'

flac.md
```diff
@@ -630,6 +630,28 @@ Decoding the coded residual thus involves selecting the right coding method, fin
...
+
+## Ogg mapping
+
+The Ogg container format is defined in [@?RFC3533]. The first packet of a logical bitstream carrying FLAC data contains the bytes `0x7F 0x46 0x4C 0x41 0x43` as defined by [@?RFC5334], followed by two bytes coding the version number of the FLAC-in-Ogg mapping. These bytes are `0x01 0x00`, meaning version 1.0 of the mapping. Following the version number bytes is a 2-byte unsigned number coded big-endian containing the number of header packets (i.e. the number of metadata blocks) that follow the first packet.  This number MAY be 0, which means the number of packets that follow is unknown. Following the 2-byte number of header packets is the `fLaC` signature and a streaminfo metadata block. Taken together this means the first packet of an logical stream of FLAC-in-Ogg is always 79 bytes. This first packet MUST NOT share a Ogg page with any other packets.
```
JeromeMartinez, line 641: I would prefer something like in https://github.com/ietf-wg-cellar/flac-specification/blob/master/flac.md#streaminfo : a table with 1 line per item, IMO it is easier to read.

flac.md
```diff
@@ -630,6 +630,28 @@ Decoding the coded residual thus involves selecting the right coding method, fin
...
+
+Following the first packet are one or more header packets i.e. metadata blocks. The first of these MUST be a vorbis comment metadata block. This is contrary to native FLAC streams, where the order of metadata blocks is not important except for the streaminfo block and where a vorbis comment metadata block is optional.
```
JeromeMartinez, line 643: Why is a vorbis comment metadata block mandatory and why the order is important?

flac.md
```diff
@@ -630,6 +630,28 @@ Decoding the coded residual thus involves selecting the right coding method, fin
...
+
+Following the header packets are audio packets. Each FLAC frame is a packet. The first FLAC audio frame must start on a new Ogg page, i.e. the last metadata block must finish its page.
```
JeromeMartinez, line 645: "Each FLAC frame is a packet" -> "a OGG packet contains 1 (and only one?) FLAC frame"

JeromeMartinez (changes requested):

flac.md
ktmf01: That is part of the definition, so I don't know why

flac.md
ktmf01: Maybe this?

```suggestion
Following the header packets are audio packets. Each audio packet contains a single FLAC frame. The first audio packet must start on a new Ogg page, i.e. the last metadata block must finish its page before any audio packets are encapsulated.
```

flac.md
JeromeMartinez: 'native FLAC' seems weird to me, because FLAC may also be "native" in a container. But maybe just me, idea can be discarded if I am the only one who think that it is weird.

flac.md
JeromeMartinez: Fine for me.

flac.md
ktmf01:
```suggestion
The Ogg container format is defined in [@?RFC3533]. The first packet of a logical bitstream carrying FLAC data is structured according to the following table

 Data    | Description
:--------|:-----------
5 byte   | Bytes `0x7F 0x46 0x4C 0x41 0x43` (as also defined by [@?RFC5334])
2 byte   | Version number of the FLAC-in-Ogg mapping. These bytes are `0x01 0x00`, meaning version 1.0 of the mapping.
2 byte   | Number of header packets (excluding the first header packet) as an unsigned number coded big-endian.
4 byte   | The `fLaC` signature
4 byte   | A metadata block header for the streaminfo block
34 byte  | A streaminfo block
 
The number of header packets MAY be 0, which means the number of packets that follow is unknown. This first packet MUST NOT share a Ogg page with any other packets. This means the first page of an logical stream of FLAC-in-Ogg is always 79 bytes. 
```

flac.md
ktmf01: Maybe simply use `unencapsulated FLAC` for the specification. That way there no need to define anything

ktmf01:
As discussed during the last virtual interim meeting, I've copied the Matroska mapping here, adding some details and added some context. I've added a reference to the MP4 mapping document and an explanation that it is too extensive to include.

Finally, I've added some comments about how changes in sample rate, bit depth and number of channels should be handled.

Please comment

flac.md
```diff
@@ -630,6 +630,51 @@ Decoding the coded residual thus involves selecting the right coding method, fin
...
 
+# Container mappings
+
+The FLAC format can be used without any container as the FLAC format already provides for a very thin transport layer. However, the functionalty of this transport is rather limited, and to be able to combine FLAC audio with video, it needs to be encapsulated by a more capable container. This presents a problem: the transport layer provided by the FLAC format mixes data that belongs to the encoded data (like blocksize and sample rate) with data that belongs to the transport (like checksum and timecode). The choice was made to encapsulate FLAC frames as they are, which means some data will be duplicated and potentially deviating between the FLAC frames and the encapsulating container.
+
+The sample rate, bit depth and number of channels MUST be the same between all frame headers, the streaminfo metadata block and (if applicable) container metadata. In case an audio stream is encoded where these parameters change at some point in the stream, this should be dealt with as specified by the container.
```
JeromeMartinez, line 637: It should be a restriction per container format, e.g. MPEG-TS can support theses changes without problem. I see that it is already documented for Matroska, so this § may be deleted and just adding a tip about that for OGG.

flac.md
```diff
@@ -630,6 +630,51 @@ Decoding the coded residual thus involves selecting the right coding method, fin
...
+
+The Matroska container format is defined in [@!I-D.ietf-cellar-matroska]. The codec ID (EBML path `\Segment\Tracks\TrackEntry\CodecID`) assigned to signal tracks carrying FLAC data is `A_FLAC` in ASCII. All FLAC data before the first audio frame (i.e. the `fLaC` ASCII signature and all metadata blocks) is stored as CodecPrivate data (EBML path `\Segment\Tracks\TrackEntry\CodecPrivate`).
+
+Each FLAC frame (including all of its subframes) is treated as a single frame in the Matroska context.
+
+In case an audio stream is encoded where audio parameters (sample rate, number of channels or bit depth) change at some point in the stream, this should be dealt with a by finishing the current Matroska segment and starting a new one with the new parameters.
+
+## ISO Base Media File Format (MP4) mapping
+
+The encapsulation definition of FLAC audio in MP4 files was deemed too extensive to include in this document. A definition document can be found at https://github.com/xiph/flac/blob/master/doc/isoflac.txt
```
JeromeMartinez, line 676: I would prefer to have at least the definition of IDs and boxes, i.e. "Overview of Track Structure" and "FLACMetadataBlock"/"FLACSpecificBox" + list of authorized FLAC blocks, in addition to the link to the full details.

ktmf01: I've added a summary, but I'm not knowledgeable enough to finish this. If anyone knows someone to contact to help out here, that would be much appreciated.

Otherwise I think a statement that audio parameters cannot change in a MP4 stream should probably be added, just like in the referenced document. I can't seem to find anything similar to Ogg's chaining and Matroska's segmenting.

flac.md
```diff
@@ -654,6 +654,57 @@ The reason for this limit is to ensure that decoders can use 32-bit integers whe
...
+
+The FLAC specific box has version number 0 and contains the metadata blocks of the FLAC stream. Contrary to the CodecPrivate data in the Matroska mapping, this excludes the `fLaC` ASCII signature.
```
JeromeMartinez, line 702:
```suggestion
The FLAC specific box extends FullBox, with version number 0 and all flags set to 0, and contains all FLAC data before the first audio frame but `fLaC` ASCII signature (i.e. all metadata blocks).
```

Avoiding to link descriptions to a different format.

flac.md
```diff
@@ -654,6 +654,57 @@ The reason for this limit is to ensure that decoders can use 32-bit integers whe
...
+
+In case an audio stream is encoded where audio parameters (sample rate, number of channels or bit depth) change at some point in the stream, this should be dealt with **TODO: Someone with knowledge of this should chime in. This is NOT handled in the mapping document**
```
JeromeMartinez, line 704:
```suggestion
In case an audio stream is encoded where audio parameters (sample rate, number of channels or bit depth) change at some point in the stream, this SHOULD be dealt in a MP4 generic manner i.e. with several `stsd` atoms and different sample-description-index values in the `stsc` atom.
```

flac.md
```diff
@@ -654,6 +654,57 @@ The reason for this limit is to ensure that decoders can use 32-bit integers whe
...
+
+Each FLAC frame is a single sample in the context of MP4 files.
```
JeromeMartinez, line 706:
```suggestion
A MP4 sample SHOULD contain one FLAC frame.
```

IIRC this form is usually seen in MP4 related docs.

JeromeMartinez (changes requested):

flac.md
JeromeMartinez: Wondering if any decoder suffers if it is out of order and/or without vorbis comment block. It may be better to avoid such constraints if not useful, IMO.

flac.md
ktmf01: Wouldn't this result in the question: what are the exceptions to this? If I remember correctly from the recent Matroska discussion, a SHOULD without documented exceptions is a MUST. What about simply

```suggestion
A MP4 sample contains one single FLAC frame.
```

flac.md
ktmf01: I see now I used should for similar phrases for Matroska and Ogg, perhaps MUST is better?

flac.md
JeromeMartinez: Technically MP4 does not limit the count of frames per MP4 samples, but all documents limiting the possibilities of MP4 indicate this limitation, as if it is something we should be afraid about. But I don't see any problem with more than 1 frame in a MP4 sample, reason I have put "SHOULD", but I am fine to remove it.

flac.md
```diff
@@ -654,6 +654,57 @@ The reason for this limit is to ensure that decoders can use 32-bit integers whe
...
+
+The sample entry code is 'fLaC'. No object type indication is registered. The channelcount and samplesize fields in the sample entry follow from the values as found in the FLAC stream. The samplerate field can be different, because FLAC can carry audio with much higher samplerates than can be coded for in the sample entry. When possible, the samplerate field should contain the sample rate as found in the FLAC stream, shifted left by 16 bits to get the 16.16 fixed point representation of the samplerate field. When the FLAC stream contains a sample rate higher than can be coded, the samplerate field contains the greatest expressible regular division of the sample rate, e.g. 48000 for sample rates of 96kHz and 192kHz or 44100 for a sample rate of 88200Hz. When the FLAC stream contain audio with an unusual samplerate that has no regular division, the maximum value of 65535.0 Hz is used. As FLAC streams with a high sample rate are common, a parser or decoder MUST read the value from the FLAC streaminfo metadata block or a frame header to determine the actual sample rate. The sample entry contains one 'FLAC specific box' with code 'dfLa'.
```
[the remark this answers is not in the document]
ktmf01:
```suggestion
The sample entry code is 'fLaC'. The channelcount and samplesize fields in the sample entry follow from the values as found in the FLAC stream. The samplerate field can be different, because FLAC can carry audio with much higher samplerates than can be coded for in the sample entry. When possible, the samplerate field should contain the sample rate as found in the FLAC stream, shifted left by 16 bits to get the 16.16 fixed point representation of the samplerate field. When the FLAC stream contains a sample rate higher than can be coded, the samplerate field contains the greatest expressible regular division of the sample rate, e.g. 48000 for sample rates of 96kHz and 192kHz or 44100 for a sample rate of 88200Hz. When the FLAC stream contain audio with an unusual samplerate that has no regular division, the maximum value of 65535.0 Hz is used. As FLAC streams with a high sample rate are common, a parser or decoder MUST read the value from the FLAC streaminfo metadata block or a frame header to determine the actual sample rate. The sample entry contains one 'FLAC specific box' with code 'dfLa'.
```

flac.md
JeromeMartinez: Reluctant with MUST, e.g in case there is a new method invented in MP4 in the future, FLAC does not care actually, it is just out of the scope. So replacing "i.e." by "e.g." (meaning for me that FLAC is not specific to how it is done in MP4).:

```suggestion
In case an audio stream is encoded where audio parameters (sample rate, number of channels or bit depth) change at some point in the stream, this MUST be dealt in a MP4 generic manner e.g. with several `stsd` atoms and different sample-description-index values in the `stsc` atom.
```

flac.md
ktmf01: One problem with FLAC is that the length of a frame is not explicitly stored anywhere. I can imagine that with PCM audio, where a frame is generally only a few bytes and of fixed length, storing several frames in a sample makes sense, but I think it needlessly complicates demuxing.

Also, this bit of text is in my view only meant to be a summary of the full document, which states the following
```
            A FLAC sample is exactly one FLAC audio FRAME (as defined
            in the FLAC[3] file specification) belonging to a FLAC
            bitstreams.
```

flac.md
ktmf01: I need to do some digging to find out whether there is any specific reason for this

flac.md
JeromeMartinez: I am usually reluctant to limit a container format, but to be honest 1 frame per sample is a classic rule, so fine for me.

flac.md
ktmf01: It seems the reasoning for this is here: https://lists.xiph.org/pipermail/flac-dev/2004-July/001567.html

In retrospect this reasoning seems a little weak: it is so the vendor string has a fixed offset from the start of the Ogg file and a decoder can use this to figure out whether it can decode a particular FLAC file. It seems the vendor string was being used as a version string instead, which it isn't of course.

> what is your recommendation here?  is it that the fLaC and
first STREAMINFO (which holds sample rate, #samples,
resolution, etc) packets be flushed together so that they
are in the same page, page 0?  the total size of both packets
is 42 bytes so this seems to be no problem.
> 
> but adding in the codec version to the first page.... in
FLAC the codec version is in the vendor tag of the vorbis
comments.  the vorbis comments are in their own metadata
block/packet, but there is no requirement for the vorbis
comments to follow the STREAMINFO immediately.  there may
be other metadata in between.  now, I can enforce it in
libOggFLAC that it follows STREAMINFO immediately, but then
the question is, how can you guarantee that the whole vorbis
comment packet will also fit in page 0, given that there is
not much restriction on the size of comments?  is it enough
that the first part of the vorbis comment packet that contains
the vendor string is in page 0?

This might have worked back in 2004, but we have many more FLAC encoders these days, and they don't follow the same versioning as libFLAC.

So, that seems to be the why. As for whether this results in problems if omitted, that is a much harder question to answer.

I agree that useless constraints should be removed, but I don't know what we're risking with this. It doesn't really seem to hurt, it is simply a minor inconvenience to muxers. I see no reason not to leave it in.

FFMPEG does it too
https://github.com/FFmpeg/FFmpeg/blob/ed83a3a5bddf4c209157dc9f041eda0721b4c3e0/libavformat/oggenc.c#L328-L329

JeromeMartinez: > I see no reason not to leave it in.

It seems weird to me that a VORBIS comment is mandatory, i.e. it is legitimate not to have one (it is optional in FLAC native stream, why not in OGG?). Maybe a SHOULD there, but not blocking for me, just annoyed by "useless" constraints.

ktmf01: Then perhaps something like:
```

Comment (JeromeMartinez) [reply flac.md]: Fine for me.

File: flac.md [hunk 11]
```diff
@@ -654,6 +654,57 @@ The reason for this limit is to ensure that decoders can use 32-bit integers whe
 
 Following the last subframe is the frame footer. If the last subframe is not byte aligned (i.e. the bits required to store all subframes put together are not divisible by 8), zero bits are added until byte alignment is reached. Following this is a 16-bit CRC, initialized with 0, with polynomial x^16 + x^15 + x^2 + x^0. This CRC covers the whole frame excluding the 16-bit CRC, including the sync code.
 
+# Container mappings
+
+The FLAC format can be used without any container as the FLAC format already provides for a very thin transport layer. However, the functionalty of this transport is rather limited, and to be able to combine FLAC audio with video, it needs to be encapsulated by a more capable container. This presents a problem: the transport layer provided by the FLAC format mixes data that belongs to the encoded data (like blocksize and sample rate) with data that belongs to the transport (like checksum and timecode). The choice was made to encapsulate FLAC frames as they are, which means some data will be duplicated and potentially deviating between the FLAC frames and the encapsulating container.
+
```
retokromer, line 660: typo: `functionalty` -> `functionality`

retokromer (approved):

JeromeMartinez (approved):

PR #162, opened 2022-08-06
Clarify use of seektable in containers

### LINKED ISSUE #160: Add note on use of seektable metadata block in container
[taken from #112]

I think clarification is needed. However I'm not sure about "that is serves no purpose at all":
* It may become relevant after demuxing. So it should be made clear if the seekpoint values refer to muxed or unmuxed byte offsets (physical offsets or per-logic stream offsets).
* Containers only ever *add* overhead. So the offsets are still valid in a "there or later" sense. So they can still be helpful for seeking as they allow an area of the file being skipped. If the demuxer/decoder has additional details it might even use a more precise guess. e.g.: "the correct position should be at the given offset plus 5%".
* In containers that store the data non-interleaved such as RIFF/WAVE the offsets are byte perfect valid as-is.

I think that a simple clarification would allow for all the above and keep being compatible without any downsides. It could read something similar to this: The offsets are within the FLAC bitstream. Any overhead added by containers is not included. Therefore when used inside containers the offsets provide a minimal offset in the physical bitstream which can aid seeking even when used within containers. It is save to restore seektables as-is when a) demuxing to native FLAC, b) remuxing into a different container.

Side note: Offsets are stored relative to the first audio block. So even if the container stores the headers in a different section (e.g. in the tail of the file, or provide them externally (think SIP/SDP)) the offsets are still valid.

PS: Maybe this discussion should go into it's own ticket.

_Originally posted by @phschafft in https://github.com/ietf-wg-cellar/flac-specification/issues/112#issuecomment-1139695702_

ktmf01:
This closes #160 

@phschafft what do you think of this wording?

### DIFF
--- a/flac.md
+++ b/flac.md
@@ -239,14 +239,15 @@ Data     | Description
 
 ## Seektable
 
-The seektable metadata block can be used to store seek points. It is possible to seek to any given sample in a FLAC stream without a seek table, but the delay can be unpredictable since the bitrate MAY vary widely within a stream. By adding seek points to a stream, this delay can be significantly reduced. Each seek point takes 18 bytes, so a seek table with 1% resolution within a stream adds less than 2KB of data. There MUST NOT be more than one seektable metadata block in a stream, but the table can have any number of seek points. There is also a special 'placeholder' seekpoint which will be ignored by decoders but which can be used to reserve space for future seek point insertion.
+The seektable metadata block can be used to store seek points. It is possible to seek to any given sample in a FLAC stream without a seek table, but the delay can be unpredictable since the bitrate may vary widely within a stream. By adding seek points to a stream, this delay can be significantly reduced.  There MUST NOT be more than one seektable metadata block in a stream, but the table can have any number of seek points.
+
+Each seek point takes 18 bytes, so a seek table with 1% resolution within a stream adds less than 2KB of data. The number of seek points is implied by the metadata header 'length' field, i.e. equal to length / 18. There is also a special 'placeholder' seekpoint which will be ignored by decoders but which can be used to reserve space for future seek point insertion.
 
 Data         | Description
 :------------|:-----------
 `SEEKPOINT`+ | One or more seek points.
 
-NOTE
-- The number of seek points is implied by the metadata header 'length' field, i.e. equal to length / 18.
+A seektable is generally not usable for seeking in a FLAC file embedded in a container, as such containers usually interleave FLAC data with other data and the offsets used in seekpoints are those of an unmuxed FLAC stream. Also, containers often provide their own seeking methods. It is however possible to store the seektable in the container along with other metadata when muxing a FLAC file, so this stored seektable can be restored on demuxing the FLAC stream into a standalone FLAC file.
 
 ### Seekpoint
 Data     | Description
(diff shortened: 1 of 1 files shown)


retokromer:
ping @ phschafft

flac.md
```diff
@@ -239,14 +239,15 @@ Data     | Description
...
+The seektable metadata block can be used to store seek points. It is possible to seek to any given sample in a FLAC stream without a seek table, but the delay can be unpredictable since the bitrate may vary widely within a stream. By adding seek points to a stream, this delay can be significantly reduced.  There MUST NOT be more than one seektable metadata block in a stream, but the table can have any number of seek points.
+
+Each seek point takes 18 bytes, so a seek table with 1% resolution within a stream adds less than 2KB of data. The number of seek points is implied by the metadata header 'length' field, i.e. equal to length / 18. There is also a special 'placeholder' seekpoint which will be ignored by decoders but which can be used to reserve space for future seek point insertion.
```
retokromer, line 244: nit: I would add a space at `2 KB`

retokromer (approved):

flac.md
ktmf01: In #167 I spelled these out (megabyte, gigabyte). Probably best to keep it consistent. As the document has a lot of mentions of bytes, perhaps best to spell this one out? What do you think?

retokromer: Yes, let’s keep it consistent.

ktmf01, line 244:
```suggestion
Each seek point takes 18 bytes, so a seek table with 1% resolution within a stream adds less than 2 kilobyte of data. The number of seek points is implied by the metadata header 'length' field, i.e. equal to length / 18. There is also a special 'placeholder' seekpoint which will be ignored by decoders but which can be used to reserve space for future seek point insertion.
```

JeromeMartinez (approved):
Just a typo, look for a double space "  ".

flac.md
```diff
@@ -239,14 +239,15 @@ Data     | Description
 
 ## Seektable
 
-The seektable metadata block can be used to store seek points. It is possible to seek to any given sample in a FLAC stream without a seek table, but the delay can be unpredictable since the bitrate MAY vary widely within a stream. By adding seek points to a stream, this delay can be significantly reduced. Each seek point takes 18 bytes, so a seek table with 1% resolution within a stream adds less than 2KB of data. There MUST NOT be more than one seektable metadata block in a stream, but the table can have any number of seek points. There is also a special 'placeholder' seekpoint which will be ignored by decoders but which can be used to reserve space for future seek point insertion.
+The seektable metadata block can be used to store seek points. It is possible to seek to any given sample in a FLAC stream without a seek table, but the delay can be unpredictable since the bitrate may vary widely within a stream. By adding seek points to a stream, this delay can be significantly reduced.  There MUST NOT be more than one seektable metadata block in a stream, but the table can have any number of seek points.
```
ktmf01, line 242:
```suggestion
The seektable metadata block can be used to store seek points. It is possible to seek to any given sample in a FLAC stream without a seek table, but the delay can be unpredictable since the bitrate may vary widely within a stream. By adding seek points to a stream, this delay can be significantly reduced. There MUST NOT be more than one seektable metadata block in a stream, but the table can have any number of seek points.
```

Issue #159, opened 2022-05-27
Limits of the PICTURE block


phschafft:
Quoting from the current flac.md:
> Note that while the length fields for MIME type, description and picture data are 4 bytes in length and could in theory code for a size up > to 4 GiB, the total metadata block size cannot exceed what can be described by the metadata block header, i.e. 16 MiB.

Based on discussions on IRC on security considerations for decoding I would suggest to "soft" limit the length of both fields as per specs. By "soft" limit I mean a clarification like "encoders SHOULD NOT generate" and "decoders MAY consider invalid" if the lengths are not within limitations. I would also like to remind everyone that a flac picture block is also used in other standards such as embedded in Vorbis Comments (e.g. for Opus) and even in ID3 (APIC frame). The 16MB limitation may/does not apply there.

I suggest to define the "soft" limit not too small but within what is considered small for modern memory management. E.g. a value of 1^(12..16)-1 bytes would work fine here and should not create any real world limitation but helping against possible attacks.

ktmf01:
> I would also like to remind everyone that a flac picture block is also used in other standards such as embedded in Vorbis Comments (e.g. for Opus) and even in ID3 (APIC frame). The 16MB limitation may/does not apply there.

Are you sure Opus refers to the FLAC specification for this? ID3 doesn't, the APIC frame is actually the inspiration for the FLAC picture metadata block structure, so it is actually the other way around. The fact that both share many similarities does not mean that limits in one apply to the other.

> I suggest to define the "soft" limit not too small but within what is considered small for modern memory management. E.g. a value of 1^(12..16)-1 bytes would work fine here and should not create any real world limitation but helping against possible attacks.

This would invalidate the majority of FLAC files with embedded album art. 2^16 byte is 65KiB which isn't sufficient for image files. From FLAC files I've bought from online stores, embedded album art files are often between 100KB and 2MB in size.

phschafft:
> > I would also like to remind everyone that a flac picture block is also used in other standards such as embedded in Vorbis Comments (e.g. for Opus) and even in ID3 (APIC frame). The 16MB limitation may/does not apply there.
> 
> Are you sure Opus refers to the FLAC specification for this?

I'm not aware of a formal specification. However comments with the key "METADATA_BLOCK_PICTURE" containing a base64 encoded picture block are well supported (e.g. opusenc) and are something found in the wild. I'm also very sure I have seen that in Ogg/Vorbis streams as well.

[...]

PR #167, opened 2022-08-19
Add FLAC specifics to security considerations

ktmf01:
I've tried to improve the security considerations section by naming a few issues I've encountered in libFLAC the last few years. Perhaps the security considerations can be improved further taking those of EBML as inspiration.

https://www.rfc-editor.org/rfc/rfc8794.html#name-security-considerations

I could also review all CVE's against libFLAC or other FLAC parsers, but I think they will mostly turn up out-of-bounds access issues.


ktmf01:
Also, I think this is sufficient to fix #159. @phschafft could you comment? There are no specific suggestions for limits here, as I found it hard to decide on that, but it is specifically mentioned

flac.md
```diff
@@ -656,7 +656,18 @@ A list of other implementations and an overview of which parts of the format the
...
+
+When metadata is handled, it is RECOMMENDED to either thoroughly test handling of extreme cases or impose reasonable limits beyond the limits of this specification document. For example, a single Vorbis comment metadata block can contain millions of valid fields. It is unlikely such a limit is ever reached except in a potentially malicious file. Likewise the media type and description of a picture metadata block can be millions of characters long, despite there being no reasonable use of such contents. One possible use case for very long character strings is in lyrics, which can be stored in Vorbis comment metadata block fields.
+
+Various kinds of metadata blocks contain length fields or fields counts. While reading a block following these lengths or counts, a decoder MUST make sure higher-level lenghts or counts (most importantly the length field of the metadata block itself) are not exceeded. As some of these length fields code string lengths, memory for which must be allocated, it is RECOMMENDED to first verify that a block is valid before allocating memory based on its contents.
```
retokromer, line 665: typo: higher-level `lenghts` or counts > higher-level `lengths` or counts

flac.md
```diff
@@ -656,7 +656,18 @@ A list of other implementations and an overview of which parts of the format the
...
+
+As with all compression algorithms, both encoding and decoding can produce an output much larger than the input. In case of decoding, the most extreme possible case of this is a frame with eight constant subframes of blocksize 65535 and coding for 32-bit PCM. This frame is only 49 byte in size, but codes for more than 2 megabyte of uncompressed PCM data. For encoding, it is possible to have an even larger size increase, although such behavior is generally considered faulty. This happens if the encoder chooses a rice parameter that does not fit with the residual that has to be encoded. In such a case, very long unary coded symbols can appear, in the most extreme case more than 4 gigabyte per sample. It is RECOMMENDED decoder and encoder implementations take precautions to prevent excessive resource utilisation in such cases.
```
retokromer, line 661: `utilisation` (UK) > `utilization` (US)

flac.md
ktmf01:
```suggestion
Various kinds of metadata blocks contain length fields or fields counts. While reading a block following these lengths or counts, a decoder MUST make sure higher-level lengths or counts (most importantly the length field of the metadata block itself) are not exceeded. As some of these length fields code string lengths, memory for which must be allocated, it is RECOMMENDED to first verify that a block is valid before allocating memory based on its contents.
```

flac.md
ktmf01:
```suggestion
As with all compression algorithms, both encoding and decoding can produce an output much larger than the input. In case of decoding, the most extreme possible case of this is a frame with eight constant subframes of blocksize 65535 and coding for 32-bit PCM. This frame is only 49 byte in size, but codes for more than 2 megabyte of uncompressed PCM data. For encoding, it is possible to have an even larger size increase, although such behavior is generally considered faulty. This happens if the encoder chooses a rice parameter that does not fit with the residual that has to be encoded. In such a case, very long unary coded symbols can appear, in the most extreme case more than 4 gigabyte per sample. It is RECOMMENDED decoder and encoder implementations take precautions to prevent excessive resource utilization in such cases.
```

retokromer (approved):

JeromeMartinez (approved):

PR #172, opened 2022-09-08
Add 32 bps to bit depth table

ktmf01:
I was under the impression this was already in, but I just noticed it wasn't.

### DIFF
--- a/flac.md
+++ b/flac.md
@@ -518,7 +518,7 @@ Value   | Bit depth
 0b100   | 16 bits per sample
 0b101   | 20 bits per sample
 0b110   | 24 bits per sample
-0b111   | reserved
+0b111   | 32 bits per sample
 
 The next bit is reserved and MUST be zero.
 


JeromeMartinez:
It makes sense but in practice is it supported in at least libflac?
I was thinking that 32-bit support is possible only with streaminfo.

ktmf01:
libFLAC 1.4.0 is planned to be released tomorrow, it will support this. I stumbled upon it while preparing the release. ffmpeg patches have been around for a while but not yet merged.

32-bit could also have been signalled through streaminfo alone, but the problem with that is that ffmpeg would have tried to decode and failed. With this set in the header, ffmpeg (without the patches) doesn't even try, which I think is a good thing. Perhaps other decoders will display similar behaviour: trying to decode a stream with 32 bps set in streaminfo and failing in unexpected ways, or fail cleanly with this until-now-reserved bit pattern.

JeromeMartinez:
> libFLAC 1.4.0 is planned to be released tomorrow, it will support this.

Great!
So we have the reference supporting it.

> 32 bps could also have been signalled through streaminfo alone, but the problem with that is that ffmpeg would have tried to decode and failed.

So I guess that this would be an issue with other bit depths signaled only with streaminfo. Need a fix for that :). It may also be a good conformance test file.

JeromeMartinez (approved):

ktmf01:
> So I guess that this would be an issue with other bit depths signaled only with streaminfo. Need a fix for that :).

Yes, but is seems to me the use of bit depths like 25 bit or 31 bit will be small compared to the use of 32 bit, which will in turn be small compared to 24 or 16 bit. Also, there is no way to retroactively patch existing ffmpeg deployments of course.

retokromer (approved):

PR #175, opened 2022-09-28
Consistency of block size, frame size, sample rate, bit depth

ktmf01:
This fixes https://github.com/ietf-wg-cellar/flac-specification/issues/139

The bit for the MP4 mapping might be tricky. Apparently MP4 sample entries have a samplerate field. So in this text samplerate refers to the MP4 field and sample rate is used when not talking about that field specifically. I don't know whether that is the best way to do this.


JeromeMartinez:
> So in this text samplerate refers to the MP4 field and sample rate is used when not talking about that field specifically.

Correct way to do it IMO.

JeromeMartinez (approved):

retokromer (approved):

PR #187, opened 2022-10-05
Warn for problems when changing audio properties mid-stream

ktmf01:
As suggested in https://hydrogenaud.io/index.php/topic,123069.0.html


flac.md
```diff
@@ -449,6 +449,8 @@ Directly after the last metadata block, one or more frames follow. Each frame co
...
 
+Note that storing audio with changing audio parameters in FLAC results in various practical problems. For example, these changes of audio parameters must happen on a frame boundary, or the process will not be lossless. When a variable blocksize is chosen to accomodate this, note that blocks smaller than 16 samples are not allowed and it is therefore not possible to store an audio stream in which these parameters change within 16 samples of the last change or the start of the file. Also, since the streaminfo metadata block can only accomodate a single set of parameters, it is only valid for part of such an the audio stream. Instead, it is RECOMMENDED to store an audio stream with changing parameters in FLAC encapsulated in a container capable of handling such changes, as these do not suffer from the mentioned limitations. See [section container mappings](#container-mappings) for details.
+
```
retokromer, line 453: `accomodate` > `accommodate`

ktmf01: I just noticed I use audio parameters as well as audio properties. I think I should change this to audio properties throughout the document. Any thoughts?

flac.md
```diff
@@ -449,6 +449,8 @@ Directly after the last metadata block, one or more frames follow. Each frame co
...
 
+Note that storing audio with changing audio parameters in FLAC results in various practical problems. For example, these changes of audio parameters must happen on a frame boundary, or the process will not be lossless. When a variable blocksize is chosen to accomodate this, note that blocks smaller than 16 samples are not allowed and it is therefore not possible to store an audio stream in which these parameters change within 16 samples of the last change or the start of the file. Also, since the streaminfo metadata block can only accomodate a single set of parameters, it is only valid for part of such an the audio stream. Instead, it is RECOMMENDED to store an audio stream with changing parameters in FLAC encapsulated in a container capable of handling such changes, as these do not suffer from the mentioned limitations. See [section container mappings](#container-mappings) for details.
```
ktmf01, line 452:
```suggestion
Note that storing audio with changing audio parameters in FLAC results in various practical problems. For example, these changes of audio parameters must happen on a frame boundary, or the process will not be lossless. When a variable block size is chosen to accommodate this, note that blocks smaller than 16 samples are not allowed and it is therefore not possible to store an audio stream in which these parameters change within 16 samples of the last change or the start of the file. Also, since the streaminfo metadata block can only accommodate a single set of parameters, it is only valid for part of such an the audio stream. Instead, it is RECOMMENDED to store an audio stream with changing parameters in FLAC encapsulated in a container capable of handling such changes, as these do not suffer from the mentioned limitations. See [section container mappings](#container-mappings) for details.
```

flac.md
retokromer: Indeed, it’s a good idea to unify.

PR #189, opened 2022-10-05
Add handling of picture metadata block

ktmf01:
As discussed on #xiph IRC

### DIFF
--- a/flac.md
+++ b/flac.md
@@ -397,17 +397,17 @@ For CD-DA, an track index point number of 0 corresponds to the track pre-gap. Th
 
 ## Picture
 
-The picture metadata block contains image data of a picture in some way belonging to the audio contained in the FLAC file. Its format is derived from the APIC frame in the ID3v2 specification. However, contrary to the APIC frame in ID3v2, the MIME type and description are prepended with a 4-byte length field instead of being null delimited strings. A FLAC file MAY contain one or more picture metadata blocks.
+The picture metadata block contains image data of a picture in some way belonging to the audio contained in the FLAC file. Its format is derived from the APIC frame in the ID3v2 specification. However, contrary to the APIC frame in ID3v2, the media type and description are prepended with a 4-byte length field instead of being null delimited strings. A FLAC file MAY contain one or more picture metadata blocks.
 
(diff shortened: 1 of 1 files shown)


flac.md
```diff
@@ -443,6 +445,13 @@ Value | Picture type
 19    | Band or artist logotype
 20    | Publisher or studio logotype
 
+In case not a picture but an URI is contained in this block, the following points apply:
+
+- In case an URI does not include a protocol specifier, the URI is considered relative to the FLAC file being processed. Such URIs are retrieved with the same protocol as the FLAC file being processed, if the protocol has a concept of relative URIs.
```
JeromeMartinez, line 450:
```suggestion
- The URI can be either in absolute or in relative form. In case an URI is in relative form, it is related to the URI of the FLAC content processed.
```

See [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986), (no protocols, but schemes; URI is already defined with both concepts and name them absolute and relative, see abstract).

Also the concept of "file" may not exist (real time broadcasting etc).

flac.md
```diff
@@ -443,6 +445,13 @@ Value | Picture type
...
+- Applications MUST obtain explicit user approval to retrieve images via remote protocols, or when retrieving a file that is not in the directory of the FLAC file being processed.
```
JeromeMartinez, line 451: Maybe to be moved in the security part?

flac.md
```diff
@@ -443,6 +445,13 @@ Value | Picture type
...
+- Applications supporting linked images SHOULD handle unavailability of URIs gracefully. They MAY report unavailability to the user.
+- Applications MAY reject processing linked images for any reason, in particular for security or privacy reasons.
```
JeromeMartinez, line 453:
```suggestion
- Applications MAY reject processing URIs for any reason, in particular for security or privacy reasons.
```

flac.md
ktmf01: I think it would be nicer to leave it here, because in the Security considerations it would be missing context I think. Also, this is the only place where the protocol specifies an URL.

On the other hand, Vorbis comments can also contain links, and some applications could be using certain tags to link to extra content. In that case the consideration would be more general indeed.

ktmf01: I've now done both, there's a more generic version added to the security considerations.

retokromer (approved):

PR #191, opened 2022-10-09
Add past change on negative LPC shift

ktmf01:
As discussed in https://mailarchive.ietf.org/arch/msg/cellar/Qe71zdfR_nsNJVfUqAzQPoXxdoY/

### DIFF
--- a/rfc_backmatter.md
+++ b/rfc_backmatter.md
@@ -79,7 +79,7 @@ One significant addition to the format was the residual coding method using a 5-
 
 ## Restriction of LPC shift to non-negative values
 
-As stated in section [linear predictor subframe](#linear-predictor-subframe), the predictor right shift is a number signed two's complement, which MUST NOT be negative. This is because right shifting a number by a negative amount is undefined behavior in the C programming language standard. The intended behavior was that a positive number would be a right shift and a negative number a left shift. It turned out that a left shift would only very rarely be of any use. Instead of implementing code to handle left shifts, the FLAC reference encoder was changed in 2007 to not generate files with negative shifts.
+As stated in section [linear predictor subframe](#linear-predictor-subframe), the predictor right shift is a number signed two's complement, which MUST NOT be negative. This is because right shifting a number by a negative amount is undefined behavior in the C programming language standard. The intended behavior was that a positive number would be a right shift and a negative number a left shift. The FLAC reference encoder was changed in 2007 to not generate LPC subframes with a negative predictor right shift, as it turned out that the use of such subframes would only very rarely provide any benefit and the decoders that were already widely in use at that point were not able to handle such subframes.
 
 # Interoperability considerations
 
--- a/rfc_backmatter.md
+++ b/rfc_backmatter.md
@@ -79,7 +79,7 @@ One significant addition to the format was the residual coding method using a 5-
 
 ## Restriction of LPC shift to non-negative values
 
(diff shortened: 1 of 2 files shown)


rfc_backmatter.md
```diff
@@ -77,6 +77,10 @@ Another change to the specification was deemed necessary during standardization
 
 One significant addition to the format was the residual coding method using a 5-bit Rice parameter. Prior to publication of this addition in July 2007, there was only one residual coding method specified, a partitioned Rice code with a 4-bit Rice parameter. The range offered by this proved too small when encoding 24-bit PCM, therefore a second residual coding method was specified identical to the first but with a 5-bit Rice parameter.
 
+## Restriction of LPC shift to non-negative values
+
+As stated in section [linear predictor subframe](#linear-predictor-subframe) the predictor right shift is a number signed two's complement, which MUST NOT be negative. This is because right shifting a number by a negative amount is undefined behavior in the C programming language standard. The intended behavior was that a positive number would be a right shift and a negative number a left shift. It turned out that a left shift would only very rarely be of any use. Instead of implementing code to handle left shifts, the FLAC reference encoder was changed in 2007 to not generate files with negative shifts.
```
retokromer, line 82: nit: I would add a comma after `As stated in section [linear predictor **subframe](#linear-predictor-subframe)`**

retokromer (approved):
one nit, otherwise LGTM

rfc_backmatter.md
ktmf01, line 82:
```suggestion
As stated in section [linear predictor subframe](#linear-predictor-subframe), the predictor right shift is a number signed two's complement, which MUST NOT be negative. This is because right shifting a number by a negative amount is undefined behavior in the C programming language standard. The intended behavior was that a positive number would be a right shift and a negative number a left shift. It turned out that a left shift would only very rarely be of any use. Instead of implementing code to handle left shifts, the FLAC reference encoder was changed in 2007 to not generate files with negative shifts.
```

rfc_backmatter.md
```diff
@@ -77,6 +77,10 @@ Another change to the specification was deemed necessary during standardization
 
 One significant addition to the format was the residual coding method using a 5-bit Rice parameter. Prior to publication of this addition in July 2007, there was only one residual coding method specified, a partitioned Rice code with a 4-bit Rice parameter. The range offered by this proved too small when encoding 24-bit PCM, therefore a second residual coding method was specified identical to the first but with a 5-bit Rice parameter.
 
+## Restriction of LPC shift to non-negative values
+
+As stated in section [linear predictor subframe](#linear-predictor-subframe), the predictor right shift is a number signed two's complement, which MUST NOT be negative. This is because right shifting a number by a negative amount is undefined behavior in the C programming language standard. The intended behavior was that a positive number would be a right shift and a negative number a left shift. It turned out that a left shift would only very rarely be of any use. Instead of implementing code to handle left shifts, the FLAC reference encoder was changed in 2007 to not generate files with negative shifts.
```
ktmf01, line 82:
```suggestion
As stated in section [linear predictor subframe](#linear-predictor-subframe), the predictor right shift is a number signed two's complement, which MUST NOT be negative. This is because right shifting a number by a negative amount is undefined behavior in the C programming language standard. The intended behavior was that a positive number would be a right shift and a negative number a left shift. The FLAC reference encoder was changed in 2007 to not generate LPC subframes with a negative predictor right shift, as it turned out that the use of such subframes would only very rarely provide any benefit and the decoders that were already widely in use at that point were not able to handle such subframes.
```

retokromer (approved):

PR #192, opened 2022-11-01
Fix allowed characters in vorbis comment field name

ktmf01:


retokromer (approved):

retokromer:
@ktmf01 I can merge this one as well, I presume?

JeromeMartinez (approved):

ktmf01:
Actually I had the idea to leave these two open, so they'd be easily referenced as Spencer works through the latest draft. However, as they are relatively small changes, this is fine too. They'll show up in the draft comparison tool too after all.

PR #193, opened 2022-11-10
Correct MD5 sum calculation for non-whole byte bit depths

ktmf01:
As suggested by Josh Coalson


JeromeMartinez:
Doesn't it change the computing? I understand that with e.g. 10-bit `11 1111 0001` (decimal negative 15):
- Before (zero padding): `0000 0011 1111 0001`
- After (sign extended): `1111 1111 1111 0001`

I actually never handled content with non-whole byte bit depths, wondering what is the code in libflac.

ktmf01:
Yes, it changes the computation. The problem is that what is in the current draft is wrong :)

It does make perfect sense though, because the sign extending keeps the value the same. With zero padding negative numbers become positive, with sign-extending the number remains -15 in your example.

retokromer (approved):

PR #194, opened 2023-01-16
Add notes on uncommon block size limitations

ktmf01:


retokromer (approved):
Good catch!

flac.md
```diff
@@ -563,7 +563,7 @@ A decoder that relies on the coded number during seeking could be vulnerable to
 
 ### Uncommon block size
 
-If the block size bits defined earlier in this section were 0b0110 or 0b0111 (uncommon block size minus 1 stored), this follows the coded number as either an 8-bit or a 16-bit unsigned number coded big-endian.
+If the block size bits defined earlier in this section were 0b0110 or 0b0111 (uncommon block size minus 1 stored), this follows the coded number as either an 8-bit or a 16-bit unsigned number coded big-endian. A value of 65535 (corresponding to a block size of 65536) is invalid and MUST NOT be used, because such a block size cannot be respresented in the streaminfo metadata block. A value from 0 up to (and including) 14, which corresponds to a block size from 1 to 15, is only valid for the last frame in a stream and MUST NOT be used for any other frame. See also [the section on the streaminfo metadata block](#streaminfo).
```
JeromeMartinez, line 566: The section on the streaminfo metadata block does not say about the last frame exception, it should be synchronized.

JeromeMartinez (changes requested):

ktmf01:
I've just tried to rephrase the streaminfo metadata block section, but I found it very hard to write something both correct and reasonably easy to comprehend. Suggestions are welcome.

flac.md
```diff
@@ -222,9 +222,9 @@ Data     | Description
...
 
-The minimum block size is excluding the last block of a FLAC file, which may be smaller. If the minimum block size is equal to the maximum block size, the file contains a fixed block size stream. Note that in case of a stream with a variable block size, the actual maximum block size MAY be smaller than the maximum block size listed in the streaminfo block, and the actual smallest block size excluding the last block MAY be larger than the minimum block size listed in the streaminfo block. This is because the encoder has to write these fields before receiving any input audio data, and cannot know beforehand what block sizes it will use, only between what bounds these will be chosen.
+FLAC specifies a minimum block size of 16 and a maximum block size of 65535. This does not apply to the block size of the last frame in a stream, because the block size of that frame must be able to accomodate for the length of the audio data the file contains. As the last block is excluded from the block sizes stated in the streaminfo block, the bit patterns corresponding to the numbers 0-15 in the minimum block size and maximum block size fields are invalid.
```
JeromeMartinez, line 225:
```suggestion
The minimum block size and the maximum block size MUST be in the 16-65535 range. The minimum block size MUST be equal to or less than the maximum block size.

Any frame but the last one MUST have a block size equal to or greater than the minimum block size and MUST have a block size equal to or lesser than the maximum block size. The last frame MUST have a block size equal to or lesser than the maximum block size, it does not have to comply to the minimum block size because the block size of that frame must be able to accommodate for the length of the audio data the stream contains. 
```

Just a suggestion, not sure it is really better, yours is also fine.

JeromeMartinez (approved):

flac.md
ktmf01: I think it is an improvement, more formal.

flac.md
ktmf01: Rephrase minimum block size context for streaminfo metadata block again

flac.md
```diff
@@ -563,7 +565,7 @@ A decoder that relies on the coded number during seeking could be vulnerable to
 
 ### Uncommon block size
 
-If the block size bits defined earlier in this section were 0b0110 or 0b0111 (uncommon block size minus 1 stored), this follows the coded number as either an 8-bit or a 16-bit unsigned number coded big-endian.
+If the block size bits defined earlier in this section were 0b0110 or 0b0111 (uncommon block size minus 1 stored), this follows the coded number as either an 8-bit or a 16-bit unsigned number coded big-endian. A value of 65535 (corresponding to a block size of 65536) is invalid and MUST NOT be used, because such a block size cannot be respresented in the streaminfo metadata block. A value from 0 up to (and including) 14, which corresponds to a block size from 1 to 15, is only valid for the last frame in a stream and MUST NOT be used for any other frame. See also [the section on the streaminfo metadata block](#streaminfo).
```
retokromer, line 568: nit: respresented > represented

ktmf01: Thanks!
```suggestion
If the block size bits defined earlier in this section were 0b0110 or 0b0111 (uncommon block size minus 1 stored), this follows the coded number as either an 8-bit or a 16-bit unsigned number coded big-endian. A value of 65535 (corresponding to a block size of 65536) is invalid and MUST NOT be used, because such a block size cannot be represented in the streaminfo metadata block. A value from 0 up to (and including) 14, which corresponds to a block size from 1 to 15, is only valid for the last frame in a stream and MUST NOT be used for any other frame. See also [the section on the streaminfo metadata block](#streaminfo).
```

PR #196, opened 2023-02-07
Clarify padding block

ktmf01:
Recently I got a question whether a padding block of length 0 was possible. As the tools readily generate such blocks, I added this to the specification. I also removed the 'n must be a multiple of 8' phrase and added a note for the same phrase for application metadata blocks

### DIFF
--- a/flac.md
+++ b/flac.md
@@ -232,11 +232,9 @@ The MD5 signature is made by performing an MD5 transformation on the samples of 
 
 ## Padding
 
-The padding metadata block allows for an arbitrary amount of padding. The contents of a padding block have no meaning. This block is useful when it is known that metadata will be edited after encoding; the user can instruct the encoder to reserve a padding block of sufficient size so that when metadata is added, it will simply overwrite the padding (which is relatively quick) instead of having to insert it into the right place in the existing file (which would normally require rewriting the entire file).
+The padding metadata block allows for an arbitrary amount of padding. This block is useful when it is known that metadata will be edited after encoding; the user can instruct the encoder to reserve a padding block of sufficient size so that when metadata is added, it will simply overwrite the padding (which is relatively quick) instead of having to insert it into the right place in the existing file (which would normally require rewriting the entire file).
 
-Data     | Description
-:--------|:-----------
-`u(n)`   | n '0' bits (n MUST be a multiple of 8)
+Note that a padding metadata block MAY have a length of 0 byte, meaning the block consists only of the header. Otherwise, the header is followed by 0x00 bytes, the number of which is defined by the block length.
 
 ## Application
 
@@ -245,7 +243,7 @@ The application metadata block is for use by third-party applications. The only 
 Data     | Description
 :--------|:-----------
 `u(32)`  | Registered application ID. (Visit the [registration page](https://xiph.org/flac/id.html) to register an ID with FLAC.)
-`u(n)`   | Application data (n MUST be a multiple of 8)
+`u(n)`   | Application data (n MUST be a multiple of 8, i.e. a whole number of bytes)
 
 ## Seektable
 


flac.md
```diff
@@ -232,11 +232,9 @@ The MD5 signature is made by performing an MD5 transformation on the samples of
...
+The padding metadata block allows for an arbitrary amount of padding. This block is useful when it is known that metadata will be edited after encoding; the user can instruct the encoder to reserve a padding block of sufficient size so that when metadata is added, it will simply overwrite the padding (which is relatively quick) instead of having to insert it into the right place in the existing file (which would normally require rewriting the entire file).
 
-Data     | Description
-:--------|:-----------
-`u(n)`   | n '0' bits (n MUST be a multiple of 8)
+Note that a padding metadata block MAY have a length of 0 byte, meaning the block consists only of the header. Otherwise, the header is followed by 0x00 bytes, the number of which is defined by the block length.
```
JeromeMartinez, line 237: Not blocking, but I have a preference for the previous syntax maybe with a more explicit link to the frame header "n '0' bits (n MUST be the size of the metadata block stored in the block header multiplied by 8)"). Same remark for the Application part.

I don't see any hint that `n` could not be 0 so I would not make it more relevant there.

ktmf01: The reason for this is this issue here: https://github.com/xiph/flac/issues/544

The issues mentioned there were that it was not clear whether having a length of 0 was allowed and that the multiple of 8 is somewhat confusing.

Maybe simply 

```
`u(n)`   | n '0' bits (n MUST be a multiple of 8, i.e. a whole number of bytes, and MAY be zero)
```

I could also leave the description for the padding block unaltered (i.e. drop this PR) and add a note to the metadata block header section that metadata block contents MUST be a whole number of bytes and that a length of 0 byte is only allowed for a padding block.

Although, in theory a length of zero is also valid for seektable blocks. I might need to look into whether that is accepted by the reference implementation

JeromeMartinez: > The reason for this is this issue here: https://github.com/xiph/flac/issues/544

Wow... Some people have a so weird interpretation of a spec... Well, if we need that for mitigating such behavior, let's go.
Same for Application actually, for mirroring the explaination.

> Maybe simply

Yes.

> I could also leave the description for the padding block unaltered (i.e. drop this PR) 

No, it is fine with it.

flac.md
```diff
@@ -232,11 +232,9 @@ The MD5 signature is made by performing an MD5 transformation on the samples of
 
 ## Padding
 
-The padding metadata block allows for an arbitrary amount of padding. The contents of a padding block have no meaning. This block is useful when it is known that metadata will be edited after encoding; the user can instruct the encoder to reserve a padding block of sufficient size so that when metadata is added, it will simply overwrite the padding (which is relatively quick) instead of having to insert it into the right place in the existing file (which would normally require rewriting the entire file).
+The padding metadata block allows for an arbitrary amount of padding. This block is useful when it is known that metadata will be edited after encoding; the user can instruct the encoder to reserve a padding block of sufficient size so that when metadata is added, it will simply overwrite the padding (which is relatively quick) instead of having to insert it into the right place in the existing file (which would normally require rewriting the entire file).
```
JeromeMartinez, line 235: Add "There MAY be one or more padding metadata blocks per FLAC stream." for clarifying the answer to another question in https://github.com/xiph/flac/issues/544 and mirroring a sentence in other parts of the spec e.g. "There MUST be no more than one streaminfo metadata block per FLAC stream".

flac.md
retokromer: >Wow... Some people have a so weird interpretation of a spec...

Indeed.

retokromer (approved):

JeromeMartinez (approved):

PR #200, opened 2023-03-09
Apply suggestions from the shepherds review for appendices

ktmf01:


rfc_backmatter.md
```diff
@@ -91,23 +91,23 @@ This appendix provides some considerations for encoder implementations aiming to
...
 
 ## Rice escape code
 
-Escapes Rice partitions are only seldom used as it turned out their use provides only very small compression improvement. As many encoders therefore do not use these by default or are not capable of producing them at all, it is likely many decoder implementation are not able to decode them correctly. Therefore, when maximum compatibility with decoders is desired it is RECOMMENDED to not use escaped Rice partitions.
+Escapes Rice partitions are only seldom used as it turned out their use provides only very small compression improvement. As many encoders therefore do not use these by default or are not capable of producing them at all, it is likely many decoder implementation are not able to decode them correctly. Therefore, maximum compatibility with decoders is archieved when FLAC files are created without any use of escaped Rice partitions.
```
retokromer, line 106: archieved > achieved

retokromer (approved):
one nit, otherwise LGTM

rfc_backmatter.md
ktmf01:
```suggestion
Escapes Rice partitions are only seldom used as it turned out their use provides only very small compression improvement. As many encoders therefore do not use these by default or are not capable of producing them at all, it is likely many decoder implementation are not able to decode them correctly. Therefore, maximum compatibility with decoders is achieved when FLAC files are created without any use of escaped Rice partitions.
```

retokromer (approved):

PR #202, opened 2023-03-13
Apply some more suggestions

ktmf01:
This should be the last PR at least until the issues sent on the mail resolve.

PR #207, opened 2023-03-16
Add references to coded number section

### LINKED ISSUE #206: Explanation of "coded number" is unclear
I found the explanation of the coded number unclear; I had to read the reference implementation to understand it. Some things that would make the specification clearer to me:
* Clarify that the decoded sample/frame number is effectively the UTF-8 code point, and that we aren't storing a textual representation of the sample/frame number.
* Give an example of how an unencoded number maps to coded bytes. Then the reader gets to see how the "extended" nature of this coding scheme actually looks. I suggest the all-ones "maximum of 36 bits unencoded".

ktmf01:
This intends to fix #206

@SpencerDawkins could you take a look at this? I'm unsure whether reviewers will take kind to part of an IETF standard being used completely out of context, but leaving it as it was might not be a good idea either.

edit: both #202 and this PR touch the same section, so merging might not be possible without some editing/updating.


ktmf01:
Ok, turns out this change doesn't really cut it. RFC2279 defines up to 6 octets, but FLAC specifies 7 octets. So, I need to change this quite a bit still.

flac.md
```diff
@@ -618,10 +618,42 @@ The next bit is reserved and MUST be zero.
...
+Following the reserved bit (starting at the fifth byte of the frame) is either a sample or a frame number, which will be referred to as the coded number. When dealing with variable block size streams, the sample number of the first sample in the frame is encoded. When the file contains a fixed block size stream, the frame number is encoded.
+
+The coded number is stored in a variable length code like UTF-8 as defined [@!RFC3629], but extended to a maximum of 36 bits unencoded, 7 byte encoded.
```
ktmf01, line 623:
```suggestion
The coded number is stored in a variable length code like UTF-8 as defined in [@!RFC3629], but extended to a maximum of 36 bits unencoded, 7 byte encoded.
```

flac.md
```diff
@@ -618,10 +618,42 @@ The next bit is reserved and MUST be zero.
...
+
+When a frame number is encoded, the value MUST NOT be larger than what fits a value 31 bits unencoded or 6 byte encoded. Please note that as most general purpose UTF-8 encoders and decoders follow [@!RFC3629], they will not be able to handle these extended codes. Furthermore, while UTF-8 is specifically used to encode characters, FLAC specifically uses it to encode numbers instead. To encode or decode a coded number, follow the procedures of section 3 of [@!RFC3629], but instead of using a character number, use a frame or sample number, and instead of the table in section 3 of [@!RFC3629], use the extended table below.
```
ktmf01, line 625:
```suggestion
When a frame number is encoded, the value MUST NOT be larger than what fits a value 31 bits unencoded or 6 byte encoded. Please note that as most general purpose UTF-8 encoders and decoders follow [@!RFC3629], they will not be able to handle these extended codes. Furthermore, while UTF-8 is specifically used to encode characters, FLAC uses it to encode numbers instead. To encode or decode a coded number, follow the procedures of section 3 of [@!RFC3629], but instead of using a character number, use a frame or sample number, and instead of the table in section 3 of [@!RFC3629], use the extended table below.
```

PR #209, opened 2023-03-25
Rename subset to streamable subset, add a few mentions

ktmf01:
As far as I know, this is the last remaining comment from the shepherds review to be addressed


H2Swine:
Yes "streamable subset" reflects the intention back then, but the "streamable subset" wording suggests that non-subset streams are _not_ streamable, and ... is that really in line with reality? Sure there are restrictions that are essential for a decoder to pick up mid-stream without file header, but is it truly instructive to suggest that -r9 or -l13 aren't "streamable"?

Of course if there is no better phrase readily available ... (maybe there is? Even a minor tweak to the phrase like "streamability subset" would reflect that it is entended to ensure, rather than this is streamable and other is not).


Oh, and line 94: "thestreamable" misses a whitespace.

ktmf01:
I just think that 'the Subset' is too vague, especially when this is mentioned in the abstract and introduction. "Streamable and low-resource usage subset" is too complicated.

One could opt for video codec like naming (H.264 has constrained baseline profile, baseline profile, main profile, extended profile etc.) but we really have two profiles, I guess that would be main and extended profile? Problem with that is that the name subset has already been used for quite a long time.

So, I'd just say that streamable subset is just a name, and the details are well explained in that section.

rfc_backmatter.md
```diff
@@ -85,13 +85,13 @@ As stated in section [linear predictor subframe](#linear-predictor-subframe), th
...
 
 This appendix provides some considerations for encoder implementations aiming to create highly compatible files. As this topic is one that might change after this document is finished, consult [this web page](https://github.com/ietf-wg-cellar/flac-specification/wiki/Interoperability-considerations) for more up-to-date information.
 
-## Non-subset streams
+## Streamable subset
 
-As described in section [format subset](#format-subset), FLAC specifies a subset of its capabilities as the Subset format. Certain decoders may choose to only decode FLAC files conforming to the limitations imposed by the Subset. Therefore, maximum compatibility with decoders is achieved when the limitations of the FLAC Subset format are followed when creating FLAC files.
+As described in section [streamable subset](#streamable-subset), FLAC specifies a subset of its capabilities as the FLAC streamable subset. Certain decoders may choose to only decode FLAC files conforming to the limitations imposed by thestreamable subset. Therefore, maximum compatibility with decoders is achieved when the limitations of the FLAC streamable subset are followed when creating FLAC files.
```
ktmf01, line 94:
```suggestion
As described in section [streamable subset](#streamable-subset), FLAC specifies a subset of its capabilities as the FLAC streamable subset. Certain decoders may choose to only decode FLAC files conforming to the limitations imposed by the streamable subset. Therefore, maximum compatibility with decoders is achieved when the limitations of the FLAC streamable subset are followed when creating FLAC files.
```

ktmf01:
See https://mailarchive.ietf.org/arch/msg/cellar/LZcR3wjn2YrNX9oOSRuBxOGYMGE/

PR #219, opened 2023-05-25
Replace use of non-inclusive terminology

ktmf01:
As [Matroska got a comment on the use of certain terms](https://mailarchive.ietf.org/arch/msg/cellar/qSjuY8fkzW6mGivfSmRciOgT034/), I went over a few lists, [this one](https://github.com/ietf/terminology/blob/main/.github/in-solidarity.yml) and [this one](https://www.nist.gov/nist-research-library/nist-technical-series-publications-author-instructions#table1) in particular.

This PR replaced the use of the words 'invalid', 'native' and 'sanity'. It also replaces the use of 'right' when not referring to a direction.

### DIFF
--- a/flac.md
+++ b/flac.md
@@ -219,7 +219,7 @@ At the start of a FLAC file or stream, following the `fLaC` ASCII file signature
 
 ## Metadata block header
 
-Each metadata block starts with a 4 byte header. The first bit in this header flags whether a metadata block is the last one, it is a 0 when other metadata blocks follow, otherwise it is a 1. The 7 remaining bits of the first header byte contain the type of the metadata block as an unsigned number between 0 and 126 according to the following table. A value of 127 (i.e. 0b1111111) is dorbidden. The three bytes that follow code for the size of the metadata block in bytes excluding the 4 header bytes as an unsigned number coded big-endian.
+Each metadata block starts with a 4 byte header. The first bit in this header flags whether a metadata block is the last one, it is a 0 when other metadata blocks follow, otherwise it is a 1. The 7 remaining bits of the first header byte contain the type of the metadata block as an unsigned number between 0 and 126 according to the following table. A value of 127 (i.e. 0b1111111) is forbidden. The three bytes that follow code for the size of the metadata block in bytes excluding the 4 header bytes as an unsigned number coded big-endian.
 
 Value   | Metadata block type
 :-------|:-----------
--- a/flac.md
+++ b/flac.md
@@ -149,7 +149,7 @@ For more information on fixed and linear predictors, see [@HPL-1999-144] and [@r
 
 If a subframe uses a predictor to approximate the audio signal, a residual is stored to 'correct' the approximation to the exact value. When an effective predictor is used, the average numerical value of the residual samples is smaller than that of the samples before prediction. While having smaller values on average, it is possible a few 'outlier' residual samples are much larger than any of the original samples. Sometimes these outliers even exceed the range the bit depth of the original audio offers.
 
-To be able to efficiently code such a stream of relatively small numbers with an occasional outlier, Rice coding (a subset of Golomb coding) is used. Depending on how small the numbers are that have to be coded, a Rice parameter is chosen. The numerical value of each residual sample is split in two parts by dividing it with `2^(Rice parameter)`, creating a quotient and a remainder. The quotient is stored in unary form, the remainder in binary form. If indeed most residual samples are close to zero and the Rice parameter is chosen right, this form of coding, a so-called variable-length code, needs less bits to store than storing the residual in unencoded form.
+To be able to efficiently code such a stream of relatively small numbers with an occasional outlier, Rice coding (a subset of Golomb coding) is used. Depending on how small the numbers are that have to be coded, a Rice parameter is chosen. The numerical value of each residual sample is split in two parts by dividing it with `2^(Rice parameter)`, creating a quotient and a remainder. The quotient is stored in unary form, the remainder in binary form. If indeed most residual samples are close to zero and a suitable Rice parameter is chosen, this form of coding, a so-called variable-length code, needs less bits to store than storing the residual in unencoded form.
 
 As Rice codes can only handle unsigned numbers, signed numbers are zigzag encoded to a so-called folded residual. For more information see section [coded residual](#coded-residual) for a more thorough explanation.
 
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
`invalid` is commented with "Taken out after causing author confusion numerous time.", it may make the document more difficult to read e.g. "tablenotvalidpatterns" seems weird, maybe good to keep "invalid".
Or replacing by "forbidden", it seems to be the actual meaning when used in this document.

ktmf01:
I've changed the 'patterns that are not valid' to 'forbidden patterns'. There are other uses of the term 'invalid' that I've left as 'not valid' as they do not refer to patterns that are explicitly forbidden but are not correct otherwise.

H2Swine:
Careful. Going from "specifies that patterns in certain places are not valid" to "specifies forbidden patterns in certain places" could possibly introduce confusion. The former is clearer that you cannot have this pattern in that place in the flac stream or file. The latter could be read as certain places _in the specification document._ What about "forbids certain patterns from occurring in specific places in a valid bitstream"?

Also there is a "dorbidden" that should be "forbidden".

(How "forbidden" relates to "MUST NOT" is beyond me. Say, whether there is a distinction between say, "a value of xx is forbidden and MUST NOT be used" and "the value MUST NOT be xx".)

ktmf01:
> Careful. Going from "specifies that patterns in certain places are not valid" to "specifies forbidden patterns in certain places" could possibly introduce confusion. The former is clearer that you cannot have this pattern in that place in the flac stream or file. The latter could be read as certain places _in the specification document._ What about "forbids certain patterns from occurring in specific places in a valid bitstream"?
> 

The whole text is the following

> Apart from reserved patterns, the format specifies forbidden patterns in certain places, meaning that the patterns MUST NOT appear in any bitstream.

So, there is already a reference to reserved patterns (which I think are pretty clear to be reserved patterns in the format from the context), it says 'the format' specifies forbidden patterns and there is the specific mention that forbidden patterns must not appear in any bitstream, which means they are in the forbidden patterns in the format. Also, it is part of the section 'format principles'. I don't see the ambiguity here?

> Also there is a "dorbidden" that should be "forbidden".

I'm pretty sure I fixed that in 29ae838

> (How "forbidden" relates to "MUST NOT" is beyond me. Say, whether there is a distinction between say, "a value of xx is forbidden and MUST NOT be used" and "the value MUST NOT be xx".)

I don't understand, please explain. It specifically states forbidden patterns MUST NOT appear in any bitstream

H2Swine:
On the former: there is a distinction between a certain place in the bitstream, and a certain place (say subsection) of the format specification. But it might be clear enough anyway, as a certain place in the specification document does correspond to parts of the bitstream anyway.

As for the latter: In a way this could be an issue with "invalid" too, but it suddenly becomes more relevant when we change phrases than when copying something that has been in there since forever. "MUST", "MUST NOT" etc is part of the language, and one should consider when introducing terminology that imposes/forbids, whether you accomplish anything you could have made forth in a clearer way by using MUST and MUST NOT directly. This is subject to the rules for use of the MUST and MUST NOT phrases in documents like these, and that is beyond my skills. But for what it is worth: at least when you are actively making changes, one should ask whether a simple "MUST NOT be xxx" is any worse than "xxx is forbidden and MUST NOT appear". I guess standard phrases like MUST NOT are there for a reason, namely to avoid competing phrases?

ktmf01:
> On the former: there is a distinction between a certain place in the bitstream, and a certain place (say subsection) of the format specification. But it might be clear enough anyway, as a certain place in the specification document does correspond to parts of the bitstream anyway.

I fail to see how one can speak of patterns in the specification document itself.

> As for the latter: In a way this could be an issue with "invalid" too, but it suddenly becomes more relevant when we change phrases than when copying something that has been in there since forever. "MUST", "MUST NOT" etc is part of the language, and one should consider when introducing terminology that imposes/forbids, whether you accomplish anything you could have made forth in a clearer way by using MUST and MUST NOT directly.

In this case, I think it is clearer to use forbidden. As Jérôme said, it conveys the actual meaning/reason.

> This is subject to the rules for use of the MUST and MUST NOT phrases in documents like these, and that is beyond my skills. But for what it is worth: at least when you are actively making changes, one should ask whether a simple "MUST NOT be xxx" is any worse than "xxx is forbidden and MUST NOT appear". I guess standard phrases like MUST NOT are there for a reason, namely to avoid competing phrases?

Of course, this is also new to me. However, it seems from previous comments on this document that these so-called BCP 14 terms are not specified as the only way, but rather, they can only be used when the convey exactly what is defined by BCP 14. You may notice that the document uses MUST NOT and must not separately, with only a slight difference in what they mean. In other words, there already 'competing terms' and as far I understand it, that is not a problem.

H2Swine:
> I fail to see how one can speak of patterns in the specification document itself.

That's not the ambiguity. Let's get back to the phrase:
"the format specifies forbidden patterns in certain places". 
What is obviously meant, is to group together "forbidden patterns in certain places [in the bitstream]". But since word order isn't that fixed, a statement of the form "the format [does something, whatever that is] in certain places" can also be read as "this specification document will in certain places [= sections, subparagraphs, sentences, tables ...] specify [something]". 

It might not be an issue. If every _"certain place"_ in the document specifies what can or can not be found in a corresponding _"certain place"_ in the bit stream, then a reader who interprets it this way should extrapolate to the other interpretation. And hopefully every valid pattern in every _"certain place"_ in the bit stream is covered by the spec too.


So ... alles gut?

flac.md
```diff
@@ -161,9 +161,9 @@ The FLAC format uses two forms of Rice coding, which only differ in the number o
...
 
-FLAC has no format version information, but it does contain reserved space in several places. Future versions of the format MAY use this reserved space safely without breaking the format of older streams. Older decoders MAY choose to abort decoding or skip data encoded using methods they do not recognize. Apart from reserved patterns, the format specifies invalid patterns in certain places, meaning that the patterns MUST NOT appear in any bitstream. These invalid patterns are usually used to make the synchronization mechanism more robust. They are listed in the following table.
+FLAC has no format version information, but it does contain reserved space in several places. Future versions of the format MAY use this reserved space safely without breaking the format of older streams. Older decoders MAY choose to abort decoding or skip data encoded using methods they do not recognize. Apart from reserved patterns, the format specifies forbidden patterns in certain places, meaning that the patterns MUST NOT appear in any bitstream. These patterns are usually defined as forbidden to make the synchronization mechanism more robust. They are listed in the following table.
```
ktmf01, line 164:
```suggestion
FLAC has no format version information, but it does contain reserved space in several places. Future versions of the format MAY use this reserved space safely without breaking the format of older streams. Older decoders MAY choose to abort decoding or skip data encoded using methods they do not recognize. Apart from reserved patterns, the format specifies forbidden patterns in certain places, meaning that the patterns MUST NOT appear in any bitstream. They are listed in the following table.
```

PR #221, opened 2023-06-05
Fix mistake in appendix A

ktmf01:
It turns out I could prove some statements wrong, so this is fixed


H2Swine:
Normally one writes `(-8)*(-8)` with parentheses when two negatives are multiplied. 
In practice it is unambiguous what is meant, so I wouldn't have bothered to nitpick had it already been in the document and not in a pull request. Fix if you think it is worth it.

rfc_backmatter.md
```diff
@@ -7,12 +7,14 @@ In order to maintain lossless behavior, all arithmetic used in encoding and deco
...
 
-In case of multiplication, the number of bits needed for the result is the size of the first variable plus the size of the second variable, but counting only one sign bit if working with signed data types. If for example a 16-bit signed integer is multiplied by a 16-bit signed integer, the result needs at least 31 bits to store without overflowing.
+In case of multiplication, the number of bits needed for the result is the size of the first variable plus the size of the second variable. If for example a 16-bit signed integer is multiplied by a 16-bit signed integer, the result needs at least 32 bits to store without overflowing. To show this in practice, the largest signed value that can be stored in 4 bit is -8. -8 * -8 is 64, which needs at least 8 bits (signed) to store.
```
ktmf01, line 17:
```suggestion
In case of multiplication, the number of bits needed for the result is the size of the first variable plus the size of the second variable. If for example a 16-bit signed integer is multiplied by a 16-bit signed integer, the result needs at least 32 bits to store without overflowing. To show this in practice, the largest signed value that can be stored in 4 bit is -8. (-8)*(-8) is 64, which needs at least 8 bits (signed) to store.
```

PR #222, opened 2023-07-03
Remove last use of 'regex' syntax

ktmf01:
Make section seektable refer to section seekpoint in the same way that cuesheet refers to cuesheet track and index point


H2Swine:
Since you have decided upon "zero or more" seekpoints (the format spec at xiph.org says one or more), then from https://github.com/xiph/flac/issues/579  it might be an idea to include a warning in the section on interoperability considerations that for max compatibility one would want to refrain from empty seektable? 
(Spawns the question: for that purpose, is a placeholder enough - or would one want a seek point number 0 at sample number 0?)

ktmf01:
I think that is a way too specific detail. It concerns the reference implementation on one specific hardware platform. I have no information about whether it breaks anything on other implementations.

PR #228, opened 2023-08-04
capitalise Vorbis

retokromer:
If I am not wrong, these are the only two occurrences of `vorbis` rather than `Vorbis`.


ktmf01:
Thanks!

PR #233, opened 2023-10-09
Attempt to address Éric Vyncke's comments

ktmf01:
See https://mailarchive.ietf.org/arch/msg/cellar/35ehSFLAkg8aGq0ddGWWYPxhN6o/

### DIFF
--- a/rfc_backmatter.md
+++ b/rfc_backmatter.md
@@ -226,13 +226,13 @@ Start  | Length   | Contents           | Description
 0x14+4 | 3 bits   | 0b001              | 2 channels
 0x14+7 | 5 bits   | 0b01111            | Sample bit depth 16
 0x15+4 | 36 bits  | 0b0000, 0x00000001 | Total no. of samples 1
-0x1a   | 16 bytes | (...)              | MD5 signature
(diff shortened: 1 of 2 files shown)


JeromeMartinez (approved):

rfc_backmatter.md
```diff
@@ -725,10 +725,10 @@ Residual  | Predictor w/o shift | Predictor | Sample value
 2         | -26  | -7  | -5
 0         | 1    | 0   | 0
 
-By lining all these samples up, we get the following input for the MD5 summing process.
+By lining all these samples up, we get the following input for the MD5 checksumming process.
```
retokromer, line 728: I would have said `MD5 checksum computing process`, but I am not a native English speaker.

retokromer (approved):

rfc_backmatter.md
ktmf01: Well, neither am I...

It has a mention here: https://en.wiktionary.org/wiki/checksum#English but it does not seem to be frequently used. Probably better to change it indeed.

ktmf01:
```suggestion
By lining all these samples up, we get the following input for the MD5 checksum calculation process.
```

PR #234, opened 2023-10-13
Attempt to address most of the editorial comments of Robert Sparks

ktmf01:
See https://mailarchive.ietf.org/arch/msg/cellar/r7kx9ixRa52lnYK4ZV8JDo8hBSg/

### DIFF
--- a/rfc_backmatter.md
+++ b/rfc_backmatter.md
@@ -3,7 +3,7 @@
 
 # Numerical considerations
 
-In order to maintain lossless behavior, all arithmetic used in encoding and decoding sample values MUST be done with integer data types to eliminate the possibility of introducing rounding errors associated with floating-point arithmetic. Use of floating-point representations in analysis (e.g., finding a good predictor or Rice parameter) is not a concern, as long as the process of using the found predictor and Rice parameter to encode audio samples is implemented with only integer math.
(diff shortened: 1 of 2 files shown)


rfc_backmatter.md
```diff
@@ -506,7 +506,7 @@ Residual  | Sample value
 -267      | -6299
 134       | -6165
 
-With this, the	 decoding of the first subframe is complete. The decoding of the second subframe is very similar, as it also uses a fixed predictor of order 1, so this is left as an exercise for the reader, the results are in the next table. The next step is undoing stereo decorrelation, which is done in the following table. As the stereo decorrelation is right-side, in which the actual ordering of the subframes is side-right, the samples in the right channel come directly from the second subframe, while the samples in the left channel are found by adding the values of both subframes for each sample.
+With this, the	 decoding of the first subframe is complete. The decoding of the second subframe is very similar, as it also uses a fixed predictor of order 1, so this is left as an exercise for the reader, the results are in the next table. The next step is undoing stereo decorrelation, which is done in the following table. As the stereo decorrelation is side-right, the samples in the right channel come directly from the second subframe, while the samples in the left channel are found by adding the values of both subframes for each sample.
```
JeromeMartinez, line 509: Nit: the extra tab between "the" and " decoding" could be removed while changing other stuff here.

ktmf01:
```suggestion
With this, the decoding of the first subframe is complete. The decoding of the second subframe is very similar, as it also uses a fixed predictor of order 1, so this is left as an exercise for the reader, the results are in the next table. The next step is undoing stereo decorrelation, which is done in the following table. As the stereo decorrelation is side-right, the samples in the right channel come directly from the second subframe, while the samples in the left channel are found by adding the values of both subframes for each sample.
```

JeromeMartinez (approved):

retokromer (approved):

PR #236, opened 2023-11-15
Move application metadata block ID registry to IANA

ktmf01:
Took a while, but here it is. @SpencerDawkins could you review this?

### DIFF
--- a/flac.md
+++ b/flac.md
@@ -309,8 +309,10 @@ The application metadata block is for use by third-party applications. The only 
 
 Data     | Description
 :--------|:-----------
-`u(32)`  | Registered application ID. (Visit the [registration page](https://xiph.org/flac/id.html) to register an ID with FLAC.)
+`u(32)`  | Registered application ID.
 `u(n)`   | Application data (n MUST be a multiple of 8, i.e., a whole number of bytes) n is 8 times the size described in the metadata block header, minus the 32 bits already used for the application ID.
+
+Application ID are registered with the IANA, see [#application-id-registry].
 
 ## Seektable
 
@@ -929,7 +931,7 @@ FLAC files may contain executable code, although the FLAC format is not designed
 
 # IANA Considerations
 
-This document registers one new media type, "audio/flac", as defined in the following section.
+This document registers one new media type, "audio/flac", as defined in the following section, and creates a new IANA registry.
 
 ## Media type registration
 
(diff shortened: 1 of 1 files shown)


mcr (approved):

retokromer (approved):

flac.md
```diff
@@ -987,6 +989,58 @@ Change controller: Internet Engineering Task Force
 Provisional registration? (standards tree only): NO
 ```
 
+## Application ID Registry
+
+This document creates a new IANA registry called the "FLAC Application Metadata Block ID" registry. The values correspond to the 32-bit identifier described in (#application).
+
+To register a new Application ID in this registry, one needs an Application ID, a description, optionally a reference to a document describing the Application ID and a Change Controller (IETF or email of registrant). The Application IDs are to be allocated according to the "First Come First Served" policy [RFC8126]. An Application ID can be any 32-bit value, but is often composed of 4 ASCII characters, to be human-readable.
+
+The FLAC Application Metadata Block ID registry is assigned the following initial values.
```
SpencerDawkins, line 998: @ktmf01, could you help me with a couple of things here? 

My understanding is that we want application IDs to be registered if they are in use, because FLAC audio files encoded using any of these application IDs can pop up at an arbitrary point in the future, and it's better for someone who needs to decode the audio file to know where to start looking - this is why FCFS is the correct choice for the registry policy. 

My understanding is that we are taking over a list of application IDs from https://xiph.org/flac/id.html, that hasn't been maintained, and many of the links on that page return 404s, etc, so although XIPH did have a reference for BUGS when it was registered, we can't provide a reference for BUGS for anyone who wants to decode BUGS audio files stored in FLAC is that correct?

Assuming so, for both of these, perhaps it's useful to say this as 

> The Application IDs are to be allocated according to the "First Come First Served" policy [RFC8126], so that there is no impediment to registering any Application IDs the FLAC community encounters, especially if they were used in audio files but were not registered when the audio files were encoded. 

and 

> The FLAC Application Metadata Block ID registry is assigned the following initial values, [taken from the XIPH.org registration page](https://xiph.org/flac/id.html).

Especially to @ktmf01 and @mcr - does that make sense?

flac.md
```diff
@@ -987,6 +989,58 @@ Change controller: Internet Engineering Task Force
...
+
+Application ID   | ASCII rendition (if available) | Description                        | Specification       | Change controller
+:----------------|:-------------------------------|:-----------------------------------|:--------------------|:---------
+0x41544348       | ATCH                           | FlacFile                           | [@FlacFile]         | IETF
```
SpencerDawkins, line 1002: I'm getting "no page found" for flacFile. Is that still working for you?

SpencerDawkins (approved):
@ktmf01 - you did good work. I have a couple of suggestions, but you don't need to accept them before you merge the PR.

flac.md
ktmf01: Works fine on my end. Did you take the whole url or just the last bit? This should work: https://web.archive.org/web/20071023070305/http://firestuff.org:80/flacfile/

flac.md
```diff
@@ -309,9 +309,11 @@ The application metadata block is for use by third-party applications. The only
 
 Data     | Description
 :--------|:-----------
-`u(32)`  | Registered application ID. (Visit the [registration page](https://xiph.org/flac/id.html) to register an ID with FLAC.)
+`u(32)`  | Registered application ID.
 `u(n)`   | Application data (n MUST be a multiple of 8, i.e., a whole number of bytes) n is 8 times the size described in the metadata block header, minus the 32 bits already used for the application ID.
 
+Application ID are registered with the IANA, see [#application-id-registry].
```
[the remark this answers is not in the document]
ktmf01:
```suggestion
Application IDs are registered with the IANA, see [#application-id-registry].
```

flac.md
ktmf01: > My understanding is that we are taking over a list of application IDs from https://xiph.org/flac/id.html, that hasn't been maintained, and many of the links on that page return 404s, etc, so although XIPH did have a reference for BUGS when it was registered, we can't provide a reference for BUGS for anyone who wants to decode BUGS audio files stored in FLAC is that correct?

Not necessarily. Providing a specification has always been optional when registering an ID. So, yes, Xiph did have a reference for BUGS, but not a specification. I've pruned that list down to instances where some form of specification is known to me.

For example, the reference for BUGS is http://www.neowizbugs.com/ which doesn't exist anymore. However, I'm reasonably sure that this player is related to a Korean music streaming service with the same name that still exists: https://en.wikipedia.org/wiki/Bugs!_(streaming_service) I haven't been able to find a specification, so I've not included any link.

---

Considering your suggestion, I think it makes perfect sense, I'll include it. I'll make the link an informative reference instead of just a link however.

PR #237, opened 2023-11-16
Formalize MusicBrainz reference, explain why no tags are defined

ktmf01:
This should address part of the DISCUSS here: https://mailarchive.ietf.org/arch/msg/cellar/vXmBfQOaKL_blgyJ5AUI599O-H8/

### DIFF
--- a/flac.md
+++ b/flac.md
@@ -363,13 +363,25 @@ Note that the Vorbis comment as used in Vorbis allows for on the order of 2\^64 
 
 ### Standard field names
 
-Only one standard field name is defined; the channel mask field, in (#channel-mask). In general, most FLAC playback devices and software recognize the following field names:
+Only one standard field name is defined; the channel mask field, in (#channel-mask). No other field names are defined, because the applicability of any field name is strongly tied to the content it is associated with. For example, field names useful for describing files that contain a single work of music would be unusable when labeling archived broadcasts, recordings of any kind, or a collection of music works. Even when describing a single work of music, different conventions exists depending on the kind of music: orchestral music differs from music by solo artists or bands.
+
+Despite no field names are formally defined, there is a general trend among devices and software capable of FLAC playback that are meant to play music. Most of those recognize at least the following field names:
 
 - Title: name of the current work.
 - Artist: name of the artist generally responsible for the current work. For orchestral works, this is usually the composer; otherwise, it is often the performer.
 - Album: name of the collection the current work belongs to.
 
-For a more comprehensive list of possible field names, [the list of tags used in the MusicBrainz project](http://picard-docs.musicbrainz.org/en/variables/variables.html) is recommended.
+For a more comprehensive list of possible field names suited for describing a single work of music in various genres, the list of tags used in the MusicBrainz project, see [@MusicBrainz], is suggested.
+
+<reference anchor="MusicBrainz" target="http://picard-docs.musicbrainz.org/en/variables/variables.html">
+  <front>
+    <title>Tags &amp; Variables - MusicBrainz Picard v2.10 documentation</title>
+    <author>
+      <organization>MusicBrainz</organization>
+    </author>
+    <date/>
+  </front>
+</reference>
 
 ### Channel mask
 


flac.md
```diff
@@ -363,13 +363,25 @@ Note that the Vorbis comment as used in Vorbis allows for on the order of 2\^64
...
+Despite no field names are formally defined, there is a general trend among devices and software capable of FLAC playback that are meant to play music. Most of those recognize at least the following field names:
 
 - Title: name of the current work.
 - Artist: name of the artist generally responsible for the current work. For orchestral works, this is usually the composer; otherwise, it is often the performer.
 - Album: name of the collection the current work belongs to.
 
-For a more comprehensive list of possible field names, [the list of tags used in the MusicBrainz project](http://picard-docs.musicbrainz.org/en/variables/variables.html) is recommended.
+For a more comprehensive list of possible field names suited for describing a single work of music in various genres, the list of tags used in the MusicBrainz project, see [@MusicBrainz], is suggested.
+
+<reference anchor="MusicBrainz" target="http://picard-docs.musicbrainz.org/en/variables/variables.html">
```
retokromer, line 376: `http://` > `https://`

ktmf01:
```suggestion
<reference anchor="MusicBrainz" target="https://picard-docs.musicbrainz.org/en/variables/variables.html">
```

retokromer (approved):

flac.md
```diff
@@ -363,13 +363,25 @@ Note that the Vorbis comment as used in Vorbis allows for on the order of 2\^64
 
 ### Standard field names
 
-Only one standard field name is defined; the channel mask field, in (#channel-mask). In general, most FLAC playback devices and software recognize the following field names:
+Only one standard field name is defined: the channel mask field, in (#channel-mask). No other field names are defined because the applicability of any field name is strongly tied to the content it is associated with. For example, field names useful for describing files that contain a single work of music would be unusable when labeling archived broadcasts, recordings of any kind, or a collection of music works. Even when describing a single work of music, different conventions exist depending on the kind of music: orchestral music differs from music by solo artists or bands.
```
JeromeMartinez, line 366:
```suggestion
Only one standard field name is defined: the channel mask field, in (#channel-mask). Other field names are free-form field names with their meaning being outside the scope of this specification because the applicability of any field name is strongly tied to the content it is associated with. For example, field names useful for describing files that contain a single work of music would be unusable when labeling archived broadcasts, recordings of any kind, or a collection of music works. Even when describing a single work of music, different conventions exist depending on the kind of music: orchestral music differs from music by solo artists or bands.
```

Or

```suggestion
Only one standard field name is defined: the channel mask field, in (#channel-mask). Other field names are free-form field names with their meaning being outside the scope of this specification.
```

(I am not sure that there is a need of more details here, it is just outside of this spec)

flac.md
```diff
@@ -363,13 +363,25 @@ Note that the Vorbis comment as used in Vorbis allows for on the order of 2\^64
...
+
+Despite the fact that no field names are formally defined, there is a general trend among devices and software capable of FLAC playback that are meant to play music. Most of those recognize at least the following field names:
```
JeromeMartinez, line 368:
```suggestion
Despite the fact that no field names are formally defined, there is a general trend among devices and software capable of FLAC playback that are meant to play music. An implementation MAY recognize the following field names:
```

For defining the level of RFC 2119 recommendation (here: optional)

flac.md
ktmf01: In last call review and IESG reviews the term 'hand-waiving' was used a few times, I believe at least once in this context. I think there is a need to explain why no tags are formalized: not because we are too "lazy" to define this but because there is a real reason why defining a one-size-fits-all approach for field names is impossible: use cases are too different. It is not out-of-scope per se, if it was straightforward I would have included it.

flac.md
ktmf01: As part of the shepherds review, a lot of RFC 2119 keywords were removed because they were in informative parts of the specification (like the section conceptual overview), and they don't really belong there. I think the same goes here: this sentence gives a hint on what a specific set of implementations often do, which isn't normative at all. So I think RFC 2119 keywords aren't really useful here?

H2Swine:
It is not at all clear to me whether the document deprecates the long-standing (but yes slightly annoying) use of ORGANIZATION. If it does, it should certainly be specific about it. If it doesn't ... it isn't clear either.  I think we have a timeline like this: 

1. FLAC has always used Vorbis comments, even if there are further restrictions. Everyone has thought of it as Vorbis comments, not "something that partially looks like Vorbis comments"; [https://xiph.org/flac/format.html](https://xiph.org/flac/format.html) calls it "an implementation of the [Vorbis comment specification](http://xiph.org/vorbis/doc/v-comment.html) (without the framing bit)", and that link points to that "[Vorbis]" reference.
2. Said reference lists a few standard field names. Including ORGANIZATION. As proposal, yes, but: 
3. Major players adopt at least some of these, including ORGANIZATION. (Mp3tag and foobar2000 are likely to be called "major players" as far as FLAC tagging goes.) The document says "proposed", and some of the proposals are taken on board by major players. It has become understood that this is a recommended Vorbis comment, and even if you don't use it ... understand an honour it as a publisher tag.
4. Time passes. ORGANIZATION sticks ... at least in some major corners of the FLAC world, it sticks.
5. Come this standard specification. It recommends a set of fields where I don't find ORGANIZATION, but right before it refers to the very same "[Vorbis]" reference which is (I think?) the cause of the use of that field name in FLAC.

How is that going to be read, when the "[Vorbis]" doc always has been taken to be what FLAC tags are, with the modification clarified in the old format spec? If it truly is relegated to "means nothing anymore except explaining why it got the name in the first place", it won't be read that way except by fresh developers who know nothing about the history and thus won't understand ORGANIZATION. While established applications will keep on using ORGANIZATION because well, it is still Vorbis comments isn't it? No, seriously, has something changed? Nah, the "[Vorbis]" reference is of course still there, as has been since infancy. If you want to lead applications and users in some common direction (and why else even have a section on it), then that in particular is something that should be mentioned.

ktmf01:
> It is not at all clear to me whether the document deprecates the long-standing (but yes slightly annoying) use of ORGANIZATION.

This is all new to me, I have never head of this long-standing use.

I think the document is pretty clear in that it defines absolutely no tags at all except the channel mask one. It is left completely unspecified. The reason for this is explained in this very PR.

H2Swine:
> > It is not at all clear to me whether the document deprecates the long-standing (but yes slightly annoying) use of ORGANIZATION.
> 
> This is all new to me, I have never head of this long-standing use.

- I cannot remember foobar2000 _not_ using it, over 15 years - but I am sure the dev would clarify it for you. Closest thing the changelog gets to suggesting it, is version 0.9.4.1: https://www.foobar2000.org/changelog-old
- dBpoweramp would not do the remapping, IIRC, but it would use ORGANIZATION and not MusicBrainz' suggestion. Again something I started using > 15 years ago. 
- I don't know how long Mp3 tag has done it, but here in 2009 the developer Florian states it is because the Vorbis comment recommendation, he links to it: https://community.mp3tag.de/t/vorbis-comment-mapping/8593
- 2009: Mentioned in a Winamp release. https://forums.winamp.com/forum/winamp/winamp-discussion/285547-winamp-5-572-released-5-57-revised-build-2830 
- 2010: JRiver fix, starts using ORGANIZATION: https://yabb.jriver.com/interact/index.php/topic,59363.msg401086.html#msg401086
- 2012, a thread on the Logitech forum: https://forums.slimdevices.com/forum/user-forums/3rd-party-software/89252-muso-support-thread/page6

You agree these amount to "major" players and "long-standing" as far as FLAC is concerned ... ?

Also: 
- The Hydrogenaud.io wiki is just a wiki, but it has been there for over ten years.
- FWIW, very little mention at https://wiki.xiph.org/Field_names , but at least some. (The spreadsheet was there in 2009.) 



> I think the document is pretty clear in that it defines absolutely no tags at all except the channel mask one. 

Not defines, but recommends ... well "recommended" out, "suggested" in, but it is still a recommendation, and because ORGANIZATION - annoying as it is IMHO but still - has for long been seen as part of the Vorbis comment implementation, then what?

ktmf01:
I don't see the problem. Anyone who wishes to do so can use this organization tag, or a label tag, or both, that is the whole point of having the field names free form. Just because it isn't mentioned in the suggested tags does not exclude it.

mcr:
If I'm understanding this issue correctly, the OP would like the document to
standardize more of the Vorbis tags.  That definitely seems out of scope for
this document.

PR #245, opened 2024-09-22
Fix typo

ledoge:


ktmf01:
Currently the document is undergoing intense scrutiny as part by the RFC editor, so I won't make any changes. They'll probably notice this as well, but thanks for bringing it up anyway.

H2Swine:
@ktmf01 : They will likely have spottted the following, but in case not ...  Page references to https://www.ietf.org/archive/id/draft-ietf-cellar-flac-14.txt : 

- Two instances: page 53 and 69 uncapitalized "vorbis". All the other (except in URL) have capital V
- page 41: "least-sigificant" should be "least-significant" 
- page 8: a "the the" that should only be a single "the"
- Also but maybe good enough, page 3: "at at". It is "looking back at at most 32", which does make sense as "looking back at" "at most 32", but "looking back at most 32" is less puzzling to the reading eye.

ktmf01:
> but in case not ...  

@H2Swine Then now is too late.

Here is the current state: https://www.rfc-editor.org/authors/rfc9639.html

All those things have been corrected AFAIK. Publication will probably be in about 3 weeks.

PR #246, opened 2025-02-08
Remove all outdated stuff

ktmf01:
As there have been a lot of changes between draft-ietf-cellar-flac-14 and RFC 9639, I think there are two possible options: update the document here, or remove it. As git preserves history anyway, I think removing is the better option, to avoid any possible conflicts with RFC 9639.

RFC 9639 links here for the example files, so these should be preserved. Also, the license file is kept and the readme is updated.

Please comment.


H2Swine:
The proposed new Readme.md says 

> Currently, it only contains the example files used in the document.

But it will also still contain a wiki at https://github.com/ietf-wg-cellar/flac-specification/wiki , and https://github.com/ietf-wg-cellar/flac-specification/wiki/Interoperability-considerations is in the RFC's reference list, and then I guess it has to stay.

ktmf01:
Ah, yes, indeed. Technically not part of the git repository but of the github 'accessories', but still useful the mention.

retokromer (approved):

PR #247, opened 2025-04-09
fix two typos

retokromer:
I stumbled upon two typos.


ktmf01:
Hi Reto,

Both issues have been addressed by the RFC editor prior to publication. This repository has not been updated to match the published RFC. As per #246 I'd like to remove flac.md anyway.

